PR #10, opened 2024-08-29
feat: calculator tool

michael-desmond:
Calculator tool based on [mathjs](https://mathjs.org/)


src/tools/calculator.ts
```diff
@@ -0,0 +1,45 @@
...
+Only use the calculator tool if you need to perform a calculation.`;
+    inputSchema() {
+      return z.object({
+        expression: z.string().describe(`The mathematical expression to evaluate (e.g., "2 + 3 * 4"). Use Math.js basic expression syntax. Constants only.`),
+      });
+    }
+  
+    constructor({...options }: BaseToolOptions = {}) {
+      super(options);
+    }
```
Tomas2D, line 39: constructor not needed

src/tools/calculator.ts
```diff
@@ -0,0 +1,45 @@
...
+    BaseToolOptions,
+    BaseToolRunOptions,
+    StringToolOutput,
+    Tool,
+    ToolInput,
+  } from "@/tools/base.js";
+  import { z } from "zod";
+  import { evaluate } from 'mathjs'
+  
+  export class CalculatorTool extends Tool<StringToolOutput, BaseToolOptions> {
```
Tomas2D, line 27: `BaseToolOptions` not needed

Tomas2D (changes requested):
Great work! 🚀

Please rebase against main + add a simple test case.

src/tools/calculator.ts
```diff
@@ -0,0 +1,56 @@
...
+  
+    // extra (has functional impact)
+    'evaluate':   function () { throw new Error('Function evaluate is disabled') },
+    'parse':      function () { throw new Error('Function parse is disabled') },
+    'simplify':   function () { throw new Error('Function simplify is disabled') },
+    'derivative': function () { throw new Error('Function derivative is disabled') },
+    'resolve':    function () { throw new Error('Function resolve is disabled') },
+  }, { override: true })
+
+  export class CalculatorTool extends Tool<StringToolOutput> {
```
matoushavlena, line 42: I would include a disclaimer/warning that are are potential security and stability risks involved in using `.evaluate`, even with the limited version. Pointing to the mathjs [documentation](https://mathjs.org/docs/expressions/security.html).

@Tomas2D This is likely not the first time we would be introducing a potentially dangerous tool. Any specific guidance on how approach this (comment at the top, having a specific flag turned on/off)? For example check the request tool from [langchain.](https://github.com/langchain-ai/langchain/blob/34fc00aff17d1913a1cfe4b205e732982f98f7b8/libs/community/langchain_community/tools/requests/tool.py#L36). I think users should know that this is not safe tool unless run in isolation. Another approach could be to add `unsafe` suffix to the tool name for all tools we cannot guarantee safeness OOTB.

matoushavlena (commented):
@Tomas2D Is there any guidance on using Web Workers for tools that might have impact on CPU/memory?

src/tools/calculator.ts
```diff
@@ -0,0 +1,56 @@
...
+    'createUnit': function () { throw new Error('Function createUnit is disabled') },
+    'reviver':    function () { throw new Error('Function reviver is disabled') },
+  
+    // extra (has functional impact)
+    'evaluate':   function () { throw new Error('Function evaluate is disabled') },
+    'parse':      function () { throw new Error('Function parse is disabled') },
+    'simplify':   function () { throw new Error('Function simplify is disabled') },
+    'derivative': function () { throw new Error('Function derivative is disabled') },
+    'resolve':    function () { throw new Error('Function resolve is disabled') },
+  }, { override: true })
```
Tomas2D, line 40: From my point of view, this should all be in the class constructor with the ability to pass the user's custom config (initialization).

src/tools/calculator.ts
Tomas2D: I agree that there should be a disclaimer.

Tomas2D: If we decide to use the flag, we would need to add it to other tools, like `WebCrawlerTool`.

src/tools/calculator.ts
```diff
@@ -0,0 +1,56 @@
...
+    name = "Calculator";
+    description = `A calculator tool that performs basic arithmetic operations like addition, subtraction, multiplication, and division. 
+Only use the calculator tool if you need to perform a calculation.`;
+    inputSchema() {
+      return z.object({
+        expression: z.string().describe(`The mathematical expression to evaluate (e.g., "2 + 3 * 4"). Use Mathjs basic expression syntax. Constants only.`),
+      });
+    }
+  
+    protected async _run({ expression }: ToolInput<this>, options?: BaseToolRunOptions) {
```
Tomas2D, line 52: remove the `options` parameter as it is not used

Tomas2D (changes requested):

src/tools/calculator.ts
michael-desmond: I added a warning re. security and stability of the tool, linking to the mathjs documentation.

src/tools/calculator.ts
michael-desmond: I moved this to the class constructor. I wasn't quite sure what you mean by users custom config, perhaps you can say more on that.

src/tools/calculator.ts
michael-desmond: Removed.

src/tools/calculator.ts
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+import {
+    BaseToolOptions,
+    StringToolOutput,
+    Tool,
+    ToolInput,
+  } from "@/tools/base.js";
+  import { z } from "zod";
+  import { create, all, MathJsInstance } from 'mathjs'
```
Tomas2D, line 24: 'MathJsInstance' is defined but never used

src/tools/calculator.ts
```diff
@@ -0,0 +1,70 @@
...
+   * https://mathjs.org/docs/expressions/security.html before using this tool. 
+   */
+  export class CalculatorTool extends Tool<StringToolOutput> {
+    name = "Calculator";
+    description = `A calculator tool that performs basic arithmetic operations like addition, subtraction, multiplication, and division. 
+Only use the calculator tool if you need to perform a calculation.`;
+
+    inputSchema() {
+      return z.object({
+        expression: z.string().describe(`The mathematical expression to evaluate (e.g., "2 + 3 * 4"). Use Mathjs basic expression syntax. Constants only.`),
```
Tomas2D, line 40: change to 
```
z.string().min(1).describe(`The mathematical expression to evaluate (e.g., "2 + 3 * 4"). Use Mathjs basic expression syntax. Constants only.`)
```

src/tools/calculator.ts
```diff
@@ -0,0 +1,70 @@
...
+      });
+    }
+
+    protected limitedEvaluate: any;
```
Tomas2D, line 44: `protected limitedEvaluate: typeof math.evaluate;`

Tomas2D (changes requested):

src/tools/calculator.ts
```diff
@@ -0,0 +1,70 @@
...
+        'reviver':    function () { throw new Error('Function reviver is disabled') },
+      
+        // extra (has functional impact)
+        'evaluate':   function () { throw new Error('Function evaluate is disabled') },
+        'parse':      function () { throw new Error('Function parse is disabled') },
+        'simplify':   function () { throw new Error('Function simplify is disabled') },
+        'derivative': function () { throw new Error('Function derivative is disabled') },
+        'resolve':    function () { throw new Error('Function resolve is disabled') },
+      }, { override: true })
+    }
```
Tomas2D, line 64: I was thinking about something like this.

```

export interface CalculatorToolInput {
   config?: ConfigOptions
   imports?: {
      entries: ImportObject | ImportObject[],
      options?: ImportOptions
   }
}

...

    constructor({ config, imports ...options }: CalculatorToolInput = {}) {
      super(options);
      const math = create(all, config)
      this.limitedEvaluate = math.evaluate
      // Disable use potentially vulnerable functions
      math.import({
        // most important (hardly any functional impact)
        'import':     function () { throw new Error('Function import is disabled') },
        'createUnit': function () { throw new Error('Function createUnit is disabled') },
        'reviver':    function () { throw new Error('Function reviver is disabled') },

        // extra (has functional impact)
        'evaluate':   function () { throw new Error('Function evaluate is disabled') },
        'parse':      function () { throw new Error('Function parse is disabled') },
        'simplify':   function () { throw new Error('Function simplify is disabled') },
        'derivative': function () { throw new Error('Function derivative is disabled') },
        'resolve':    function () { throw new Error('Function resolve is disabled') },
         ...imports?.entries,
      }, { override: true, ...imports?.options })
    }

...
```

Tomas2D (changes requested):
CI/CD fails because of invalid formatting. Run `yarn format:fix` to resolve that issue. Then, I will merge it.

Tomas2D (approved):

PR #40, opened 2024-09-30
feat(ibm-vllm): add llm adapter

jezekra1:


src/adapters/ibm-vllm/utils/build-client.ts
```diff
@@ -0,0 +1,140 @@
...
+              options,
+            );
+            setTimeout(() => {
+              try {
+                cache.delete(modelSpecificUrl);
+                client.close();
+              } catch {
+                /* empty */
+              }
+            }, clientShutdownTimeout).unref();
```
jezekra1, line 122: This is the cleanup/caching mechanism for memory leaks ported from API, but with a few differences:
- Client is a method `Proxy`. Before, `buildClient` returned a cached version of client which destroys itself in a few minutes, this is bad if you create a global instance in your code. With the Proxy, you get a new or cached client everytime you call a method on it
- `unref()` on `setTimeout`, because otherwise it holds up the node process with an unresolved Promise
- Using different cache package `lru-cache` - because `cache-manager` we used was updated and is requiring serializable objects now, which GRPC client is not. (Although the LRU might be an overkill, you'd need to call more than 100 different models for it to work, which is impossible), should I refactor this with a `Map`?

src/adapters/ibm-vllm/utils/wrappers.ts
```diff
@@ -0,0 +1,67 @@
...
+    { signal, ...options }: CallOptions & { signal?: AbortSignal } = {},
+  ): Promise<TResponse> => {
+    return new Promise<TResponse>((resolve, reject) => {
+      const call = fn(request, options, (err, response) => {
+        signal?.removeEventListener("abort", abortHandler);
+        if (err) {
+          reject(err);
+        } else {
+          if (response === undefined) {
+            throw new Error("Invalid response");
```
Tomas2D, line 39: `throw new FrameworkError("Invalid response from GRPC server", [err]);`

Tomas2D: !!! Throwing an error inside a promise constructor yields an unhandled rejection; do this instead.

```
reject(new FrameworkError("Invalid response from GRPC server", [err]))
```

Tomas2D: Also, what about canceling the call?

src/adapters/ibm-vllm/utils/wrappers.ts
```diff
@@ -0,0 +1,67 @@
...
+  fn: (request: TRequest, options: CallOptions) => ClientReadableStream<TResponse>,
+) {
+  return async (
+    request: TRequest,
+    { signal, ...options }: CallOptions & { signal?: AbortSignal } = {},
+  ): Promise<ClientReadableStream<TResponse>> => {
+    const stream = fn(request, options);
+    const abortHandler = () => {
+      stream.cancel();
+    };
```
Tomas2D, line 62: `const abortHandler = () => stream.cancel();`

src/adapters/ibm-vllm/utils/wrappers.ts
```diff
@@ -0,0 +1,67 @@
...
+        } else {
+          if (response === undefined) {
+            throw new Error("Invalid response");
+          }
+          resolve(response);
+        }
+      });
+      const abortHandler = () => {
+        call.cancel();
+      };
```
Tomas2D, line 46: `const abortHandler = () => stream.cancel();`

src/adapters/ibm-vllm/utils/build-client.ts
```diff
@@ -0,0 +1,140 @@
...
+  ttl: DEFAULT_GRPC_CLIENT_TTL,
+});
+
+export function buildClient({
+  url = getEnv("IBM_VLLM_URL"),
+  credentials = {
+    rootCert: getEnv("IBM_VLLM_ROOT_CERT"),
+    privateKey: getEnv("IBM_VLLM_PRIVATE_KEY"),
+    certChain: getEnv("IBM_VLLM_CERT_CHAIN"),
+  },
```
Tomas2D, line 73: Consider using `parseEnv` instead to have validation in place.

src/adapters/ibm-vllm/utils/build-client.ts
Tomas2D: Using the following mechanism is not needed.

You are ready to go by adding a class-getter with a cache decorator.
If you need to do cleanup, then introduce `destroy` method on it. You can add an auto-destroy mechanism in there if you want to.

See: 
- https://github.com/i-am-bee/bee-agent-framework/blob/main/src/tools/database/sql.ts#L84-L97
- https://github.com/i-am-bee/bee-agent-framework/blob/main/src/tools/database/sql.ts#L146-L160

Tomas2D: If you would like to have a factory separately, you can use the `CacheFn` construct.

package.json
```diff
@@ -57,7 +57,7 @@
   },
   "scripts": {
     "clean": "rimraf dist",
-    "build": "yarn clean && yarn ts:check && tsup",
+    "build": "yarn clean && yarn ts:check && NODE_OPTIONS='--max-old-space-size=8192' tsup",
```
Tomas2D, line 60: This project is not that big. I noticed that the build is very fast; however, the DST generation is very slow. Could you look in the TSUP documentation to see if there is a way to speed it up?

scripts/copyright.sh
```diff
@@ -44,7 +44,7 @@ fi
 if command -v nwa &> /dev/null; then
   nwa add -l apache -c "$AUTHOR" src dist tests scripts
 elif command -v docker &> /dev/null; then
-  docker run -it -v "${PWD}:/src" ghcr.io/b1nary-gr0up/nwa:main add -l apache -c "$AUTHOR" src dist tests scripts
+  docker run --rm -v "${PWD}:/src" ghcr.io/b1nary-gr0up/nwa:main add -l apache -c "$AUTHOR" src dist tests scripts
```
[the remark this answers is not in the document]
jezekra1: This is because `-it` is "interactive" mode and tries to attach to your terminal, which does not work in husky pre-commit script:
```
the input device is not a TTY
husky - pre-commit script failed (code 1)
```

Also if you don't include `--rm`, it will leave an orphaned container on your disk every time (=on every commit).

scripts/ibm_vllm_generate_protos.sh
```diff
@@ -0,0 +1,39 @@
...
+  --"outDir=${GRPC_TYPES_PATH}" \
+  "${GRPC_PROTO_PATH}"/*.proto
+
+# Fix imports: relative -> absolute, add .js extension
+sed -i.bak -E "s| from '\.\.?([^']*)'| from '@/${GRPC_TYPES_PATH//.\/src\//}\1.js'|g" \
+  "${GRPC_TYPES_PATH}"/*.ts \
+  "${GRPC_TYPES_PATH}"/**/*.ts
+rm "${GRPC_TYPES_PATH}"/**/*.bak "${GRPC_TYPES_PATH}"/*.bak
+yarn run lint:fix "${GRPC_TYPES_PATH}"
+yarn prettier --write "${GRPC_TYPES_PATH}"
```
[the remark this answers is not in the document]
jezekra1: This is not configurable :/, they mention it in the [docs](https://github.com/grpc/proposal/blob/master/L70-node-proto-loader-type-generator.md#rationale):
> These alternatives were considered to generating files in a directory structure based on the protobuf package structure: 
> - Generate a single file with all of the types. This creates a large, unwieldy file, making it hard for a human to evaluate changes. This can also greatly increase code duplication if multiple files need to be generated.

src/adapters/ibm-vllm/utils/wrappers.ts
jezekra1: > Also, what about canceling the call?
What do you mean, there is the listener on signal `"abort"`

src/adapters/ibm-vllm/chat.ts
```diff
@@ -0,0 +1,193 @@
...
+
+  merge(other: GrpcChatLLMOutput): void {
+    Cache.getInstance(this, "messages").clear();
+    this.raw.merge(other.raw);
+  }
+
+  getTextContent(): string {
+    const [message] = this.messages;
+    return message.text;
+  }
```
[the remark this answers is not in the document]
jezekra1: This is the same in many adapters, we should refactor it in multiple places then

src/adapters/ibm-vllm/llm.ts
```diff
@@ -0,0 +1,263 @@
...
+        },
+        {
+          signal: options?.signal,
+        },
+      );
+      for await (const chunk of stream) {
+        const typedChunk = chunk as GenerationResponse__Output;
+        const { text, ...rest } = typedChunk;
+        if (text.length > 0) {
+          // TODO remove condition once repetition checker is fixed in FW
```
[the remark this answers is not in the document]
jezekra1: It is actually, the GRPC call it sometimes yields empty string in the stream:

image

It fails in our tests and I assume it's unexpected:
```
  it("Streams", async () => {
    const llm = createLLM();
    for await (const chunk of llm.stream("Hello world!")) {
      expect(chunk).toBeInstanceOf(IBMvLLMOutput);
      expect(chunk.text).toBeTruthy(); // HERE
    }
  });
```

src/adapters/ibm-vllm/client.ts
```diff
@@ -0,0 +1,218 @@
...
+    if (!this.usedDefaultCredentials) {
+      throw new ValueError(
+        "Cannot serialize a client with credentials passed directly. Use environment variables.",
+      );
+    }
+    return {
+      options: R.omit(this.options, ["credentials"]),
+    };
+  }
+  loadSnapshot(snapshot: ReturnType<typeof this.createSnapshot>) {
```
Tomas2D, line 214: newline

src/adapters/ibm-vllm/llm.ts
```diff
@@ -150,25 +143,21 @@ export class IBMvLLM extends LLM<IBMvLLMOutput, IBMvLLMGenerateOptions> {
...
         },
-        {
-          signal: options?.signal,
-        },
+        { signal: run.signal },
       );
       const output = response.responses.at(0);
       if (!output) {
-        throw new Error("Missing output");
+        throw new LLMError("Missing output");
```
Tomas2D, line 160: `throw new LLMError("Missing output", [], { context: { response }});`

src/adapters/ibm-vllm/llm.ts
```diff
@@ -25,17 +25,17 @@ import {
...
 import { Emitter } from "@/emitter/emitter.js";
-import { GenerationResponse__Output } from "@/adapters/ibm-vllm/types/fmaas/GenerationResponse.js";
+import { GenerationResponse__Output } from "@/adapters/ibm-vllm/types.js";
 import { shallowCopy } from "@/serializer/utils.js";
 import { FrameworkError, NotImplementedError } from "@/errors.js";
 import { assign } from "@/internals/helpers/object.js";
 import { ServiceError } from "@grpc/grpc-js";
 import { Client } from "@/adapters/ibm-vllm/client.js";
 import { GetRunContext } from "@/context.js";
+import { BatchedGenerationRequest } from "./types.js";
```
Tomas2D, line 38: Unify imports. They are all pointing to the same file.

Tomas2D (changes requested):

Tomas2D (approved):

PR #79, opened 2024-10-10
feat: improve Python tool

JanPokorny:
Allow running many new CLI tools and Python libraries made available through the latest Code Interpreter executor image. Plus some prompt optimization. Code Interpreter now runs Xonsh -- a syntatic extension over Python bringing shell compatibility -- which allows the LLM to write shell commands directly and subsequently not get lost in the double escaping.

### DIFF
--- a/src/tools/python/output.ts
+++ b/src/tools/python/output.ts
@@ -38,21 +38,25 @@ export class PythonToolOutput extends ToolOutput {
   }
 
   getTextContent() {
-    const fileList = this.outputFiles
-      .map((file) => `- [${file.filename}](${PythonToolOutput.FILE_PREFIX}:${file.id})`)
-      .join("\n");
-    return `The code exited with code ${this.exitCode}.
-stdout:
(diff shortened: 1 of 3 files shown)


src/tools/python/output.ts
```diff
@@ -38,21 +38,22 @@ export class PythonToolOutput extends ToolOutput {
...
-
-stderr:
-\`\`\`
-${this.stderr}
-\`\`\`
-
-${fileList ? "Files that were created or modified:\n" + fileList : "No files were created or modified."}`;
+    return (
+      (this.exitCode === 0
+        ? "The code executed successfully.\n"
```
Tomas2D, line 43: Would you like nothing to be outputted if the code has been executed successfully? I assume you should print the stdout.

Tomas2D (changes requested):
Please @matoushavlena run some test cases.

Tomas2D (approved):

PR #86, opened 2024-10-14
feat(code-interpreter): improve file input prompt

JanPokorny:
Change format of the JSON Schema in order to allow the LLM to easily comprehend what files are available and under which name

### DIFF
--- a/src/tools/python/python.ts
+++ b/src/tools/python/python.ts
@@ -28,13 +28,14 @@ import { z } from "zod";
 import { BaseLLMOutput } from "@/llms/base.js";
 import { LLM } from "@/llms/llm.js";
 import { PromptTemplate } from "@/template.js";
-import { differenceWith, isShallowEqual, isTruthy, mapToObj, unique } from "remeda";
-import { PythonStorage } from "@/tools/python/storage.js";
+import { differenceWith, isShallowEqual, isTruthy, map, unique } from "remeda";
+import { PythonFile, PythonStorage } from "@/tools/python/storage.js";
 import { PythonToolOutput } from "@/tools/python/output.js";
 import { ValidationError } from "ajv";
 import { ConnectionOptions } from "node:tls";
 import { AnySchemaLike } from "@/internals/helpers/schema.js";
 import { RunContext } from "@/context.js";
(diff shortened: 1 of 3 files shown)


src/tools/python/python.ts
```diff
@@ -76,28 +76,43 @@ export class PythonTool extends Tool<PythonToolOutput, PythonToolOptions> {
...
     const files = await this.storage.list();
+    const fileIds = unique(map(files, ({ id }) => id));
+
+    function hasAtLeastTwoElements<T>(array: readonly T[]): array is [T, T, ...T[]] {
+      return array.length > 1;
+    }
+
+    function hasAtLeastOneElement<T>(array: readonly T[]): array is [T, ...T[]] {
+      return array.length > 0;
+    }
```
Tomas2D, line 87: Create the following function in `src/internals/helpers/array.ts`

```typescript
type HasMinLength<T, N extends number, T2 extends any[] = []> =
  T2['length'] extends N
    ? [...T2, ...T[]]
    : HasMinLength<T, N, [any, ...T2]>;

export function hasMinLength<T, N extends number>(
  arr: T[],
  n: N
): arr is HasMinLength<T, N> {
  return arr.length >= n
}
```

src/tools/python/python.ts
```diff
@@ -158,14 +175,7 @@ export class PythonTool extends Tool<PythonToolOutput, PythonToolOptions> {
...
   ) {
-    const inputFiles = await this.storage.upload(
-      Object.entries(input.inputFiles ?? {})
-        .filter(([k, v]) => Boolean(k && v))
-        .map(([id, filename]) => ({
-          id,
-          filename: filename!.split(":").at(-1) ?? filename!,
-        })),
-    );
+    const inputFiles = await this.storage.upload((input.inputFiles as PythonFile[]) ?? []);
```
Tomas2D, line 178: This changes the behavior. In the prior implementation, we filtered out `falsy` values.

Tomas2D (changes requested):

src/tools/python/python.ts
JanPokorny: That's redundant, it's not allowed by the JSON schema.

PR #90, opened 2024-10-15
feat(code-interpreter): rename "hash" to "pythonId"

JanPokorny:
Rename `hash` to signify that we no longer require this to be a content hash of the file.


Tomas2D (approved):

Tomas2D:
Why not just `id`?

JanPokorny:
@Tomas2D I'm not sure whether it is a good idea to reuse `id` for the upload identifier. One thing is `pythonId` gets generated internally by Code Interpreter for new files, which might not correspond to the way the caller generates the IDs. I think it's better to keep it this way for now.

PR #91, opened 2024-10-15
feat(internals): use different client for sse

Tomas2D:


Tomas2D:
The library does not seem to be updated.

PR #97, opened 2024-10-16
fix(code-interpreter): avoid crash when no files supplied

JanPokorny:


Tomas2D (approved):

Tomas2D:
I think that the correct commit message should be
```
fix(tools): avoid crashing code interpreter when no files supplied
```

PR #98, opened 2024-10-16
fix(code-interpreter): prevent agent from misusing IDs

JanPokorny:
Agent could not understand how to (not) use the IDs, so we just don't give the IDs to the agent anymore.

This has some implications regarding uniqueness of filenames, since we essentially use them as IDs now (among the results of `storage.list()`). Having these be unique used to be a soft requirement anyway, since the agent could not differentiate between files with the same name either way. The PR is written in a way where it does not matter whether `storage.list()` returns results with non-unique filenames, it just arbitrarily picks one of the results.

We should probably follow up by either forbidding upload of non-unique filenames to a conversation, or by auto-deduplicating (like browser downloads do).

Note that we still use the ID as a part of the link in the output. This does not seem to have any negative effects so it's left as-is for now.

### DIFF
--- a/src/tools/python/python.ts
+++ b/src/tools/python/python.ts
@@ -14,13 +14,7 @@
  * limitations under the License.
  */
 
-import {
-  BaseToolOptions,
-  BaseToolRunOptions,
-  Tool,
-  ToolInput,
-  ToolInputValidationError,
-} from "@/tools/base.js";
+import { BaseToolOptions, BaseToolRunOptions, Tool, ToolInput } from "@/tools/base.js";
 import { createGrpcTransport } from "@connectrpc/connect-node";
 import { PromiseClient, createPromiseClient } from "@connectrpc/connect";
 import { CodeInterpreterService } from "bee-proto/code_interpreter/v1/code_interpreter_service_connect";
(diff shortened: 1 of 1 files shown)


src/tools/python/python.ts
```diff
@@ -74,34 +67,18 @@ export class PythonTool extends Tool<PythonToolOutput, PythonToolOptions> {
...
-    const files = await this.storage.list();
-    const fileIds = unique(map(files, ({ id }) => id));
-
-    const zodFileId = hasMinLength(files, 2)
-      ? z.union(map(files, (file) => z.literal(file.id).describe(file.filename)))
-      : hasMinLength(files, 1)
-        ? z.literal(files[0].id).describe(files[0].filename)
-        : z.undefined();
-
+    this.files = await this.storage.list();
```
Tomas2D, line 73: I think that storing results in `this.files` is unnecessary.

Tomas2D: You can enable caching on the `inputSchema` method via the `@Cache` decorator, which you then revalidate in the `_run` method. 

See: https://i-am-bee.github.io/bee-agent-framework/#/cache?id=cache-decorator-cache (Complex Example)

src/tools/python/python.ts
```diff
@@ -170,7 +126,10 @@ export class PythonTool extends Tool<PythonToolOutput, PythonToolOptions> {
     _options: BaseToolRunOptions | undefined,
     run: RunContext<this>,
   ) {
-    const inputFiles = await this.storage.upload((input.inputFiles as PythonFile[]) ?? []);
+    const uniqueFiles = uniqueBy(this.files ?? (await this.storage.list()), (f) => f.filename);
+    const chosenFileNames = (input.inputFiles ?? []) as string[];
+    const chosenFiles = uniqueFiles.filter((file) => chosenFileNames.includes(file.filename));
+    const inputFiles = await this.storage.upload(chosenFiles);
```
Tomas2D, line 132: Use `pipe` function from `remeda` package.

Tomas2D (changes requested):

src/tools/python/python.ts
JanPokorny: I need the `id`s that are deliberately _not_ part of the `inputSchema`, thus I need to cache the full `storage.list()` result. Making a new function with `@Cache` only to invalidate it would just complicate the code -- how is `@Cache({ cacheKey: SingletonCacheKeyFn })` and `Cache.getInstance(this, "id").clear()` clearer over simply using an instance variable?

Tomas2D (approved):

PR #103, opened 2024-10-21
chore: support local model for running e2e tests

mhdawson:
Add support for using a local model for e2e testing.

Currently e2e testing requires a remote service which may cost people $ to run. Add the option
to use a model running locally if OLLAMA_HOST is set. If an API key is specified the associated
service will be used in preference to OLLAMA_HOST so this change should have no effect on
people not trying to use a local model.

### Description

If OLLAMA_HOST is set and no other API keys are providced use that host to run the e2e tests.

### Checklist

- [X] I have read the [contributor guide](https://github.com/i-am-bee/bee-agent-framework/blob/main/CONTRIBUTING.md)
- [X] Linting passes: `yarn lint` or `yarn lint:fix`
- [X] Formatting is applied: `yarn format` or `yarn format:fix`
- [X] Unit tests pass: `yarn test:unit`
- [X] E2E tests pass: `yarn test:e2e`
- [X] Tests are included 
- [X] Documentation is changed or added
- [X] Commit messages and PR title follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary)

### DIFF
--- a/tests/utils/llmFactory.ts
+++ b/tests/utils/llmFactory.ts
@@ -20,6 +20,9 @@ import { BAMChatLLM } from "@/adapters/bam/chat.js";
 import { OpenAIChatLLM } from "@/adapters/openai/chat.js";
 import { WatsonXChatLLM } from "@/adapters/watsonx/chat.js";
 import { GroqChatLLM } from "@/adapters/groq/chat.js";
+import { OllamaChatLLM } from "@/adapters/ollama/chat.js";
+import { Ollama } from "ollama";
+import { Agent, Dispatcher } from "undici";
 
 export function createChatLLM(): ChatLLM<ChatLLMOutput> {
   if (process.env.GENAI_API_KEY) {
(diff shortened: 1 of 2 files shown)


mhdawson:
It takes about 10 minutes for the e2e tests to pass with my local ollama instance with a 4070ti super, but unless I have to debug issues that's better for me than having to use a service where I pay for the requests.

Tomas2D (changes requested):
The code looks fine, but I'm not sure if we should force people to use `llama3.1:70b-instruct-q5_K_M`. Do you have any ideas?

Also make the function `noTimeoutFetch`  inline so it is automatically typed.

mhdawson:
@Tomas2D we could read the model from a .env file, however, not all versions work/pass the e2e tests. That was the smallest/lowest quantization that I found worked so there is some value in sharing which model will succeed when running the e2e tests.

Tomas2D (changes requested):
Are you sure? I was able to pass all E2E tests with ollama3.1:8b

mhdawson:
@Tomas2D I'll give it a try again, with the smaller models. I had tried a few and seen failures but maybe something else was going on.

mhdawson:
Not sure what I had wrong when I ran the tests earlier, but super happy that they pass with the smaller model.

Updated to inline as requested, to use llama3.1:8b by default and to use process.env.OLLAMA_MODEL to override it exists./

Tomas2D (changes requested):
I think that `slow` version is not required. By the way you can set `timeout` in vitest config for e2e tests.

tests/utils/llmFactory.ts
```diff
@@ -35,6 +38,20 @@ export function createChatLLM(): ChatLLM<ChatLLMOutput> {
...
     });
+  } else if (process.env.OLLAMA_HOST) {
+    return new OllamaChatLLM({
+      modelId: process.env.OLLAMA_MODEL ?? "llama3.1:8b",
+      parameters: {
+        temperature: 0,
+      },
+      client: new Ollama({
+        host: process.env.OLLAMA_HOST,
+        fetch: (input: any, init: any) => {
```
Tomas2D, line 49: either add proper types or remove typing completely

mhdawson:
@Tomas2D updated as requested. I had to work around the problem that the undici RequestInit does not extend the standard fetch one  ( I guess since it does not implement all of the options) but I was using an option not defined in the standard fetch one.

mhdawson:
Rebasing to pull in the latest seem to results in failures running the e2e tests, but I also tried using OpenAI instead of ollama and got a bunch of failures as well, so I don't think that is related to this PR.

Is there a known issue with the e2e tests or something I need to do beyond rebasing after the addition of the examples to the e2e tests?

Tomas2D:
The E2E Tests and E2E Examples should be fixed now.

Tomas2D (approved):

mhdawson:
@Tomas2D thanks for confirming, rebased.

PR #108, opened 2024-10-23
feat(tools): improve PDF capabilities

JanPokorny:
The agent seems to be able to do certain PDF tasks only through certain PDF libraries and there is no silver bullet, so let's just let it use all of the common ones.

Furthermore, the agent seems to forget to import symbols (especially commonly with PyPDF2), so we give it a directive to use qualified imports which side-steps this issue.


Tomas2D (approved):
Great. However, scope of the commit should be tools

Issue #21, opened 2024-09-12
Reduce number of files in project root


matoushavlena:
**Is your feature request related to a problem? Please describe.**
The project root contains a lot of dev/configuration related files that make people need to scroll heavily before getting to the README. Also, these files usually don't convey much useful information about the project.

**Describe the solution you'd like**
Evaluate each of these files/folder and determine if they could be moved/omitted/replaced/etc.

markstur:
Most of these files belong in the root dir and it is beneficial to keep them there so they are found by CLI commands as well as IDEs (and their plugins), but I'll try to go through them and see how many can be moved or consolidated without breaking that too much.  I like to experiment with them anyway.

mmurad2:
@markstur are there pending items or this ready to close?

markstur:
There's more.  I was tempted to open this up for Hacktoberfest small chores, but need to be more specific about which files can move and where.  I'll do that and probably just finish this.

Tomas2D:
`.lintstagedrc.json` can be easily moved to `package.json`

markstur:
> `.lintstagedrc.json` can be easily moved to `package.json`

Thanks, I actually have that and some other changes that I just didn't push yet.  Will try to get those in on Monday

markstur:
Moving .lintstagedrc.json, .release-it.json, and commitlint.config.ts to package.json.  Some devs might prefer the separate files, but to support this issue these all can go in package.json.

One problem with this is the special function for ignoring dependabot in commit-lint.  I haven't found a way to move that ts code into package.json.  It does, however, look like dependabot has fixed its problem.  So I don't think the ignore is needed.

.nvmrc can be moved for devs, but it broke a CI action.  So TBD.

The .git* files should not move.
.env.template I do not think should move.
.yarnrc.yml cannot move

The *.md files and LICENSE are pretty standard in root.  I don't see anything that should move.  GitHub recognizes them.

docker-compose.yml usually belongs in root (I didn't try moving that one).

eslint.config.js could be moved (e.g. to scripts/configs) and then the commands use --config to make this work, but then devs using IDEs would not naturally pick this up so I would not move that one.

vitest.config.ts could also move if we use the config option to specify the file with running tests, but vitest then has problems referring back to tsconfig.json and package.json.  Probably don't want to move these.

timrbula:
@markstur @Tomas2D Is this issue done given the linked PR is merged or is there a bit more to do?

[...]

PR #117, opened 2024-10-28
Issue 21 nvmrc

markstur:

Part of: https://github.com/i-am-bee/bee-agent-framework/issues/21

### Description

To reduce clutter in the root dir, .nvmrc can be removed. Instead of using this file to indicate the version that we prefer, the CONTRIBUTING.md now specifies the version.

In the near future, we should bump this version to 22.x, because 20.x recently went into maintenance (22 is active LTS). In addition, with more tests we should go to full support for LTS versions, but for now we are focusing on this one which is the most tested.

### Checklist

- [ ] Tests are included 


Tomas2D:
- Firstly, we should never support even the NodeJS version as they are experimental.
- `.nvmrc` is a special file that is recognized by the `nvm` manager and gets automatically set in my environment
- if we want to force developers to use a specific NodeJS version we may want to use `devEngines` property in `package.json`
- node 22 is LTS, that is right. However, the adoptions are not that big, so I think that we can still continue with 20.x

markstur:
I didn't understand part of your comment (see below) but if you are saying let's leave .nvmrc where it is, then I agree.  Let's close this PR.

Other replies...

> * Firstly, we should never support even the NodeJS version as they are experimental.

This is the comment I don't understand.  Did you mean the odd numbered non-LTS releases. Yes, I'd skip those.

> * `.nvmrc` is a special file that is recognized by the `nvm` manager and gets automatically set in my environment

yes, personally I'd keep it but we had issue #21 which is another opinion and this is an attempt at 1 less file but you have to follow instructions.  If it breaks normal workflow for too many devs then, I'll close the PR.

> * if we want to force developers to use a specific NodeJS version we may want to use `devEngines` property in `package.json`
> * node 22 is LTS, that is right. However, the adoptions are not that big, so I think that we can still continue with 20.x

20.x has recently gone into maintenance mode (long-term), so everyone should move to 22, but you are correct that we'd be ahead of the crowd.   We e2e test with 18, 20, 22.  I think we need to do a bit more and then move to 22 "soonish", but we should support at least both 20.x and 22.x.

More importantly...  we should not use 20.15.1 like we do.  At least that needs to be like 20.x or lts/iron or we need to manually update that regularly to get important updates.  And honestly, this PR makes that preferred version hard-coded in more places which is also bad.

PR #126, opened 2024-10-31
doc: add link checker and fixup broken links

mhdawson:
Refs: https://github.com/i-am-bee/bee-agent-framework/issues/115

Closes: #

### Description

Add a link checker that can be run with `yarn docs:links`

Add [linkinator](https://www.npmjs.com/package/linkinator) as a development dependency.  License is MIT and it seems to be what pops up when you search for link checkers.

Fixup the remaining broken links that were reported when run so that it now passes.

I've not added it as something that gets run as a pre-commit check but that could make sense and it might also make sense to have it run as a PR test check. Having said that I know that for the nodejs.org website the link checking was a sometimes a bit painful in terms of failing on non-related PRs (for example when a link goes stale as opposed to a broken link being introduced).  I'd be happy to look at doing either of those two as follow on PRs if they are something the project wants.

### Checklist

- [X] I have read the [contributor guide](https://github.com/i-am-bee/bee-agent-framework/blob/main/CONTRIBUTING.md)
- [X] Formatting is applied: `yarn format` or `yarn format:fix`
- [X] Unit tests pass: `yarn test:unit`
- [X] E2E tests pass: `yarn test:e2e`
- [ ] Tests are included 
- [X] Commit messages and PR title follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary)


mhdawson:
Note, there are some pre-existing lint failures which are not related to this PR:

```
/home/midawson/agentic/bee-agent-framework/src/adapters/ibm-vllm/chatPreset.ts
  116:22  error  A record is preferred over an index signature  @typescript-eslint/consistent-indexed-object-style

/home/midawson/agentic/bee-agent-framework/src/internals/types.ts
  108:22  error  A record is preferred over an index signature  @typescript-eslint/consistent-indexed-object-style

✖ 2 problems (2 errors, 0 warnings)
  2 errors and 0 warnings potentially fixable with the `--fix` option.

```

package.json
```diff
@@ -152,6 +153,7 @@
     "js-yaml": "^4.1.0",
     "json-schema-to-typescript": "^15.0.2",
     "jsonrepair": "^3.9.0",
+    "linkinator": "^6.1.2",
```
Tomas2D, line 156: should be part of `devDependencies`

package.json
```diff
@@ -119,6 +119,7 @@
     "start:bee": "yarn start -- examples/agents/bee.ts",
     "docs:build": "embedme --source-root=. docs/**/*.md && cp *.md docs/ && yarn lint:fix docs/ && yarn prettier --write docs/",
     "docs:check": "yarn docs:build --verify",
+    "docs:links": "linkinator \"**/*.md\" --markdown --directory-listing --resurse --skip node_modules --skip \"https://ai.meta.com/blog/meta-llama-3-1/\" ",
```
Tomas2D, line 122: Please explain.

Tomas2D (changes requested):
This may be sufficient for now, but it would be great to have something that checks the final HTML. The current solution may not catch everything.

package.json
mhdawson: I should have mentioned that `https://ai.meta.com/blog/meta-llama-3-1/` is skipped because it is flagged as failing with a 400, but I could not see anything wrong with the link and I could access it ok in my browser.

I think that's likely what you were asking about, most of the other options I think reflect the name of the option with --markdown saying to check markdown files, and --directory-listing asking for directory info

package.json
mhdawson: @Tomas2D fixed

package.json
mhdawson: I removed resurse as it was a type and obviously not needed

Tomas2D: - Yes, but it seems like we could leverage the functionality of the target library to check HTML.
- I would put this script under `docs:check` script

Regarding the error. I assume you are getting it because the crawler is detected as bot, but nevermind.

mhdawson: > I would put this script under docs:check script

Do you mean add `&& yarn docs:links` to docs:check or integrate the contents of the `docs:links` target directly into `docs:checks` ?


> Yes, but it seems like we could leverage the functionality of the target library to check HTML.

@Tomas2D I can take a look to also include HTML files, are those just html files that are in the repo itself versus being generated from markdown or something like that (I've not looked to see if there is any automation/generation of a website).

I'm out all next week so it won't be until after I get back that I could look at the HTML side of things. Also happy to add in a follow on PR.

mhdawson: I added the `yarn docs:links` to the `docs:checks` target, let me know if that was not what you wanted.

Tomas2D: There is just one single HTML file (`index.html`) that loads Docsify, which converts Markdown documents into HTML. Hence, everything is done by the client (browser).

Tomas2D (approved):
Thank you for your contribution 🚀 . This is very handy.

package.json
Tomas2D: Feel free to do so in the next PR. It is just an improvement; there is, so there's no need to rush.

Issue #110, opened 2024-10-25
OpenMeteo tool optimization


michael-desmond:
**Is your feature request related to a problem? Please describe.**
Using openMeteo with Granite 3.0 8b and I ran into a few issues that may be worth us looking into. Also touches on the general topic of tool optimization.

1. The tool name `openMeteo` is not language intuitive i.e. its not obviously descriptive of its function. As such the LLM will sometimes use the tool to perform actions its not intended for. 
2. The tool responds with metadata that is probably not useful to the calling agent. 
https://github.com/i-am-bee/bee-agent-framework/blob/9e107d75f570032528f1b46dec231f8d23db8741/src/tools/weather/openMeteo.ts#L45C1-L52C21
3. Inclusion of both daily and hourly weather data seems unnecessary. A model with a 4k context window has its token memory consumed by a single response in some forecasting circumstances. 
4. The selection of current, daily and hourly data are somewhat arbitrary and the naming is not ideal

**Describe the solution you'd like**
1. Rename the tool to something intuitive such as GetWeatherForecast of WeatherForecast to make obvious its function. This should marginally help any LLM to more consistently utilize the tool.
2. Minimize the tool response, removing  latitude, longitude, generationtime_ms, utc_offset_seconds, timezone, timezone_abbreviation and elevation.
3. Add a configuration option to limit response to daily, hourly or both. This way smaller models could be configured to save token memory.
4. Do a pass over the variables selected for the various forecast ranges i.e. 
hourly-> Apparent Temperature, Relative Humidity (2 m), Precipitation Probability (temperature, humidity, precipitation)
daily -> Maximum Apparent Temperature (2 m), Minimum Apparent Temperature (2 m), Precipitation Probability Max
current -> Apparent Temperature, Relative Humidity (2 m), Precipitation

Tomas2D:
1. That is a matter of subject because `OpenMeteo` is a well-known service. Renaming is a challenging task because, in the future, you can have multiple weather forecast providers in the framework. How would you differentiate between them?
2. Yes, we can probably remove those extra metadata. Feel free to do a PR. I am just worried about the timezone, as well.
3. How would you do that?
4. 👍🏻 

What do you think @matoushavlena ?

geneknit:
@Tomas2D this is a broader issue around LLM capabilities and needing language that an LLM can traverse. LLMs shouldn't be expected to generate a plan to call OpenMeteo, they should be expected to plan to get_weather_forecast. 
[...]

PR #155, opened 2024-11-06
feat(tools): openMeteo optimization

michael-desmond:
partially fixes: #110 

### Description

- Adds `hourly` and `daily` boolean tool options that are defaulted to `true`. If set to `false` on tool construction then the hourly or daily information will be omitted from the tool result. So for a smaller context window LLM the hourly data can be switched off. 
-  Omits a set of properties from the api response that are likely not useful to the agent, these omitted properties are stored in an array so can be reintroduced easily.
- Hourly and daily properties of the tool result are now optional in the response interface definition.
- Updated the api query fields (temp, humidity etc).
- Removed the code that was adjusting the start and end timestamps and now just create the date in UTC. I believe this code was causing incorrect dates i.e.
  - The date is created in UTC midnight when the agent provides a the date in the specified format so "2024-11-06" is "2024-11-06T00:00:00.000Z" UTC. 
  - setHours is applied to local time, which sets the time to midnight on the 2024-11-05, because midnight UTC is 1900 the prior day EST. And midnight EST is 5AM UTC the same day i.e. "2024-11-05T05:00:00.000Z" UTC. 
  - The timezone offset adjustment will then correct this to midnight on the same day i.e "2024-11-05T00:00:00.000Z". So the agent is looking for weather on the 6th, but its start date is on the 5th. 
- When omiting the fields from the response object I needed to cast the result as unknown and then recast to OpenMeteoToolResponse. Seems like im doing something incorrect here.
 

### Checklist
- [x] Tests are included 

### DIFF
--- a/src/tools/weather/openMeteo.ts
+++ b/src/tools/weather/openMeteo.ts
@@ -29,13 +29,13 @@ import { Cache } from "@/cache/decoratorCache.js";
 import { RunContext } from "@/context.js";
 import { getProp, setProp } from "@/internals/helpers/object.js";
 
-export interface ResponseFilterOptions {
-  omittedResponseKeys: (keyof OpenMeteoToolResponse)[];
+export interface ResponseFilter {
+  excludedKeys: (keyof OpenMeteoToolResponse)[];
 }
 
 interface ToolOptions extends BaseToolOptions {
   apiKey?: string;
-  responseFilter?: ResponseFilterOptions;
+  responseFilter?: ResponseFilter;
 }
 
 type ToolRunOptions = BaseToolRunOptions;
@@ -111,6 +111,25 @@ export class OpenMeteoTool extends Tool<
     this.register();
   }
 
+  public constructor(options: Partial<ToolOptions> = {}) {
+    super({
+      ...options,
+      responseFilter: options?.responseFilter ?? {
+        excludedKeys: [
+          "latitude",
+          "longitude",
+          "generationtime_ms",
+          "utc_offset_seconds",
+          "timezone",
+          "timezone_abbreviation",
+          "elevation",
+          "hourly",
+          "hourly_units",
+        ],
+      },
+    });
+  }
+
   protected preprocessInput(rawInput: unknown) {
     super.preprocessInput(rawInput);
 
@@ -181,8 +200,8 @@ export class OpenMeteoTool extends Tool<
 
     let data: OpenMeteoToolResponse = await response.json();
 
-    if (this.options?.responseFilter?.omittedResponseKeys) {
-      data = omit(data, this.options.responseFilter.omittedResponseKeys);
+    if (this.options?.responseFilter?.excludedKeys) {
+      data = omit(data, this.options.responseFilter.excludedKeys);
     }
 
     return new JSONToolOutput(data);
--- a/tests/e2e/tools/openMeteo.test.ts
+++ b/tests/e2e/tools/openMeteo.test.ts
@@ -42,39 +42,37 @@ describe("OpenMeteo", () => {
 
     expect(response.isEmpty()).toBe(false);
     expect(response.result).toMatchObject({
-      timezone: "UTC",
       current: expect.any(Object),
       current_units: expect.any(Object),
-      hourly: expect.any(Object),
-      hourly_units: expect.any(Object),
       daily: expect.any(Object),
       daily_units: expect.any(Object),
     });
   });
 
-  it("Filters", async () => {
+  it("Custom filter", async () => {
     instance = new OpenMeteoTool({
       responseFilter: {
-        omittedResponseKeys: [
+        excludedKeys: [
           "latitude",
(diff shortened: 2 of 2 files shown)


src/tools/weather/openMeteo.ts
```diff
@@ -24,12 +24,16 @@ import {
...
 import { z } from "zod";
 import { createURLParams } from "@/internals/fetcher.js";
-import { isNullish, pick, pickBy } from "remeda";
+import { isNullish, omit, pick, pickBy } from "remeda";
 import { Cache } from "@/cache/decoratorCache.js";
 import { RunContext } from "@/context.js";
 import { getProp, setProp } from "@/internals/helpers/object.js";
 
-type ToolOptions = { apiKey?: string } & BaseToolOptions;
+type ToolOptions = {
```
Tomas2D, line 32: Please use `interface` instead of `type`.

src/tools/weather/openMeteo.ts
```diff
@@ -43,20 +47,23 @@ interface LocationSearch {
   language?: string;
 }
 
+const omittedKeys: string[] = [
```
Tomas2D, line 50: I would make these part of `constructor` so that a user can easily modify them without forcing us to make changes in the framework.

TIP: `string[]` -> `(keyof OpenMeteoToolResponse)[]` + add missing values to`OpenMeteoToolResponse` interface.

Tomas2D (changes requested):

src/tools/weather/openMeteo.ts
michael-desmond: Thanks for the tip. I refactored and now you can just provide a response filter based on the OpenMeteoToolResponse keys and filter anything you like.

michael-desmond:
Applied a slight redesign based on the review feedback. Instead of switching on/off `daily` and `hourly` data requests, now you can  just provide a `responseFilter` and filter out any properties from the response. 

So to have minimal response data you could do:

```
    new OpenMeteoTool({
      responseFilter: {
        omittedResponseKeys: [
          "latitude",
          "longitude",
          "timezone_abbreviation",
          "elevation",
          "utc_offset_seconds",
          "generationtime_ms",
          "hourly",
          "hourly_units",
        ],
      },
    }),
```

This is less efficient because the api call goes and gets all the data, but code is simpler.

src/tools/weather/openMeteo.ts
```diff
@@ -24,12 +24,20 @@ import {
...
 import { createURLParams } from "@/internals/fetcher.js";
-import { isNullish, pick, pickBy } from "remeda";
+import { isNullish, omit, pick, pickBy } from "remeda";
 import { Cache } from "@/cache/decoratorCache.js";
 import { RunContext } from "@/context.js";
 import { getProp, setProp } from "@/internals/helpers/object.js";
 
-type ToolOptions = { apiKey?: string } & BaseToolOptions;
+export interface ResponseFilterOptions {
+  omittedResponseKeys: (keyof OpenMeteoToolResponse)[];
```
Tomas2D, line 33: `omittedResponseKeys` -> `excludedKeys`

src/tools/weather/openMeteo.ts
```diff
@@ -24,12 +24,20 @@ import {
...
 import { z } from "zod";
 import { createURLParams } from "@/internals/fetcher.js";
-import { isNullish, pick, pickBy } from "remeda";
+import { isNullish, omit, pick, pickBy } from "remeda";
 import { Cache } from "@/cache/decoratorCache.js";
 import { RunContext } from "@/context.js";
 import { getProp, setProp } from "@/internals/helpers/object.js";
 
-type ToolOptions = { apiKey?: string } & BaseToolOptions;
+export interface ResponseFilterOptions {
```
Tomas2D, line 32: `ResponseFilterOptions` -> `ResponseFilter`

tests/e2e/tools/openMeteo.test.ts
```diff
@@ -42,15 +42,57 @@ describe("OpenMeteo", () => {
...
+        omittedResponseKeys: [
+          "latitude",
+          "longitude",
+          "timezone_abbreviation",
+          "elevation",
+          "utc_offset_seconds",
+          "generationtime_ms",
+          "timezone",
+          "hourly",
+          "hourly_units",
```
Tomas2D, line 67: You originally had these values as defaults, but you deleted them. I might keep them as default.

Tomas2D (changes requested):

src/tools/weather/openMeteo.ts
michael-desmond: Done.

tests/e2e/tools/openMeteo.test.ts
michael-desmond: I added a default response filter with these keys excluded.

Tomas2D (approved):

PR #161, opened 2024-11-08
feat(agent): use constraint decoding for tool names in Bee

Tomas2D:
- Improves speed for cases where there are no tools
- Ensures tool name is also valid
- Does not handle cases where LLM should not use any tool at all - in other words LLM will use for instance `Python` tool with `printf` instead of generating invalid function name like `None`.


matoushavlena (commented):
Let's keep this in draft until we confirm no negative impact on generation speed and quality.

matoushavlena (approved):

PR #180, opened 2024-11-18
feat(agents): add experimental Streamlit agent

Tomas2D:


src/agents/experimental/streamlit/agent.ts
```diff
@@ -0,0 +1,181 @@
...
+      content += delta;
+      await run.emitter.emit("newToken", { delta, state: { content }, chunk });
+    }
+    const result = this.parse(content);
+
+    const assistantMessage = BaseMessage.of({
+      role: Role.ASSISTANT,
+      text: content,
+    });
+    await this.memory.addMany([userMessage, assistantMessage].filter(isTruthy));
```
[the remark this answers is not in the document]
Tomas2D: This is global memory. Prepare works with local (runner) memory.

src/agents/experimental/streamlit/agent.ts
```diff
@@ -0,0 +1,181 @@
...
+
+    const runMemory = new TokenMemory({
+      llm: this.input.llm,
+      capacityThreshold: 0.85,
+      syncThreshold: 0.6,
+      handlers: {
+        removalSelector(msgs) {
+          const msgToRemove =
+            msgs.length > 3 &&
+            (msgs.find((msg) => msg.role === Role.ASSISTANT) ??
```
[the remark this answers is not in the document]
Tomas2D: Feel free to propose a better alternative. However, I think this is intended. The oldest one will be removed.

src/agents/experimental/streamlit/agent.ts
```diff
@@ -0,0 +1,181 @@
...
+    return { runMemory, userMessage };
+  }
+
+  protected parse(content: string): Result {
+    const match = findFirstPair(content, ["```python-app\n", "```\n"]);
+    return {
+      raw: content,
+      app: match ? match.inner : null,
+      text: match
+        ? `${content.substring(0, match.start)}${content.substring(match.end + 1)}`
```
JanPokorny, line 167: This expects only one python-app block to be present, but from my experience the agent sometimes writes multiple code blocks with the last one usually being the one you want to use.

JanPokorny: Also the information where the code block was is lost in `text`. It would probably work better if we returned a structure like

```
[
    { type: "text", content: "..." },
    { type: "app", content: "..." },
    { type: "text", content: "..." },
    // ...
]
```

Then the consumer could use the code blocks as they see fit.

examples/agents/experimental/streamlit.ts
```diff
@@ -0,0 +1,32 @@
...
+    const response = await agent.run({ prompt }).observe((emitter) => {
+      emitter.on("newToken", (data) => {
+        reader.write(`Agent (token received) 🤖 : `, data.delta);
+      });
+    });
+
+    reader.write(`Agent (text) 🤖 : `, response.result.text);
+    if (response.result.app) {
+      reader.write(`Agent (app) 🤖 : `, response.result.app);
+    }
```
pilartomas, line 28: This is quite unwieldy when the consumer structures the code mostly around emitter (like bee-api does). Would it be possible to have an event emitted or at least make the parse method public?

src/agents/experimental/streamlit/agent.ts
```diff
@@ -0,0 +1,182 @@
...
+
+    let content = "";
+    for await (const chunk of this.input.llm.stream(runMemory.messages, {
+      signal: run.signal,
+    })) {
+      const delta = chunk.getTextContent();
+      content += delta;
+      await run.emitter.emit("newToken", { delta, state: { content }, chunk });
+    }
+    const result = this.parse(content);
```
pilartomas, line 105: This is quite unwieldy when the consumer structures the code mostly around emitter (like bee-api does). Would it be possible to have an event emitted or at least make the parse method public?

Tomas2D: There is no reason to make `parse` public. You would not get any beneficial data during streaming, and once executions end, you will get the parsed (and raw) result.

src/agents/experimental/streamlit/agent.ts
Tomas2D: Updated.

src/agents/experimental/streamlit/agent.ts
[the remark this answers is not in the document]
jezekra1: oops, I see now

jezekra1 (approved):

PR #193, opened 2024-11-22
feat(agents): generalize BeeAgent

Tomas2D:
Generalize the `Runner` concept in BeeAgent, allowing us to adapt other models (like Granite 3.0).

Breaking Changes:
- Remove `GraniteBeeAgent` as it is now part of `BeeAgent` (if the Granite model is provided).
- Unify some types (moved to a single file)
- Simplify Granite implementation
- Use one-liner JSON for tool schema for Granite instead of the beautified version (cc @michael-desmond)
- refactor core Bee agent run method

Ref: #178


michael-desmond:
A note on the tool schema from granite. The granite chat template indicates that the `available_tools` should be formatted as pretty json with indent=4, with a newline between each tool json. I would suggest that we maintain this formatting to be as accurate as possible to granite spec.

```
 "chat_template": "{%- if tools %}\n    {{- '<|start_of_role|>available_tools<|end_of_role|>\n' }}\n    {%- for tool in tools %}\n    {{- tool | tojson(indent=4) }}\n    {%- if not loop.last %}\n        {{- '\n\n' }}\n    {%- endif %}\n    {%- endfor %}\n    {{- '<|end_of_text|>\n' }}\n{%- endif %}\n{%- for message in messages %}\n    {%- if message['role'] == 'system' %}\n    {{- '<|start_of_role|>system<|end_of_role|>' + message['content'] + '<|end_of_text|>\n' }}\n    {%- elif message['role'] == 'user' %}\n    {{- '<|start_of_role|>user<|end_of_role|>' + message['content'] + '<|end_of_text|>\n' }}\n    {%- elif message['role'] == 'assistant' %}\n    {{- '<|start_of_role|>assistant<|end_of_role|>'  + message['content'] + '<|end_of_text|>\n' }}\n    {%- elif message['role'] == 'assistant_tool_call' %}\n    {{- '<|start_of_role|>assistant<|end_of_role|><|tool_call|>' + message['content'] + '<|end_of_text|>\n' }}\n    {%- elif message['role'] == 'tool_response' %}\n    {{- '<|start_of_role|>tool_response<|end_of_role|>' + message['content'] + '<|end_of_text|>\n' }}\n    {%- endif %}\n    {%- if loop.last and add_generation_prompt %}\n    {{- '<|start_of_role|>assistant<|end_of_role|>' }}\n    {%- endif %}\n{%- endfor %}",
```

[Granite tokenizer config](https://huggingface.co/ibm-granite/granite-3.0-8b-instruct/blob/main/tokenizer_config.json)

src/agents/bee/runners/default/runner.ts
```diff
@@ -443,16 +221,178 @@ export class BeeAgentRunner extends Serializable {
...
+        variables: {
+          tools: async () => {
+            return await Promise.all(
+              this.input.tools.map(async (tool) => ({
+                name: tool.name,
+                description: tool.description.replaceAll("\n", ".").replace(/\.$/, "").concat("."),
+                schema: JSON.stringify(
+                  await tool.getInputJsonSchema(),
+                  (() => {
+                    const ignoredKeys = new Set(["minLength", "maxLength", "$schema"]);
```
jezekra1, line 249: Nitpick: This works recursively right? It's not equivalent to the `R.omit` that granite runner used before, but it's probably better this way.

jezekra1 (approved):

src/agents/bee/runners/default/runner.ts
Tomas2D: Yes, this works recursively. Hence, it is not equivalent to the original implementation. However, it is a fix of the original implementation because it never really worked (the schema object typically doesn't have such keywords at a top level (maybe just `$schema`)).

PR #204, opened 2024-11-30
feat(agents): retry on a parser error

Tomas2D:
Retry if LLM generates something that does not adhere to the communication structure defined in the system prompt.

TODOs:
- improves the "error" prompt
- consider using another "role"
- consider whether to remove the message after successful generation or not

### DIFF
--- a/src/agents/bee/prompts.ts
+++ b/src/agents/bee/prompts.ts
@@ -186,3 +186,9 @@ export const BeeToolNotFoundPrompt = new PromptTemplate({
 Use one of the following functions: {{#trim}}{{#tools}}{{name}},{{/tools}}{{/trim}}
 {{/tools.length}}`,
 });
+
+export const BeeSchemaErrorPrompt = new PromptTemplate({
+  schema: z.object({}).passthrough(),
+  template: `Error: The generated response does not adhere to the communication structure mentioned in the system prompt.
+You communicate only in instruction lines. Valid instruction lines are 'Thought' followed by either 'Function Name' + 'Function Input' + 'Function Output' or 'Final Answer'.`,
+});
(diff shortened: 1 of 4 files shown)


src/agents/bee/prompts.ts
```diff
@@ -186,3 +186,9 @@ export const BeeToolNotFoundPrompt = new PromptTemplate({
 Use one of the following functions: {{#trim}}{{#tools}}{{name}},{{/tools}}{{/trim}}
 {{/tools.length}}`,
 });
+
+export const BeeSchemaErrorPrompt = new PromptTemplate({
+  schema: z.object({}).passthrough(),
+  template: `Error: The generated response does not adhere to the communication structure mentioned in the system prompt.
+You communicate only in instruction lines. Valid instruction lines are 'Thought' followed by either 'Function Name' + 'Function Input' + 'Function Output' or 'Final Answer'.`,
```
jezekra1, line 193: Idea (to test in prompt lab): Should we repeat the exact instructions from the system prompt?

How about extracting the "communication structure" section of the system prompt to a separate variable similar to `instructions` and using it here too?

jezekra1 (approved):

src/agents/bee/prompts.ts
Tomas2D: The problem is that the communication structure is dynamic. The presence of tools influences it.

See

```
{{/tools.length}}
# Communication structure
You communicate only in instruction lines. The format is: "Instruction: expected output". You must only use these instruction lines and must not enter empty lines or anything else between instruction lines.
{{#tools.length}}
You must skip the instruction lines Function Name, Function Input and Function Output if no function calling is required.
{{/tools.length}}
```

So I am not sure whether it is worth it to introduce a new system, `PromptTemplate`, that is going to be used as part of `BeeSystemPrompt.` I would keep it as is.

PR #207, opened 2024-12-02
fix(adapters): vLLM JSON Schema guided decoding

JanPokorny:
The parameter `json_schema` is supposed to be a string, not object.


Tomas2D (approved):
Good catch!

Issue #148, opened 2024-11-05
Improve ability to customize Bee for a given use case 


mmurad2:
**Is your feature request related to a problem? Please describe.**
The current CLI experience gives the impression that Bee works OTTB. While it does for simple tasks, for more complex use cases we learned that narrow bees with additional instructions on only the min required tools works best.

**Describe the solution you'd like**
Improve CLI experience to suggest specifying a role and tools to be used.

**Describe alternatives you've considered**
Open to alternatives. We could consider exposing additional optional settings for user to define.

**Additional context**
Would be good to align more closely to UI experience. UI is changing "instructions" to "role" FYI.

mmurad2:
Scope subject to change depending on current discussion. Feel free to update ticket description.

Tomas2D:
This issue should ideally leverage the functionality in the following module https://github.com/i-am-bee/bee-agent-framework/blob/main/examples/helpers/io.ts.
Would you like to work on this? @psschwei

psschwei:
Yeah, I can take this one

psschwei:
Based on comments in #211 , I don't think my help is wanted here.

Tomas2D:
Stale

PR #211, opened 2024-12-02
docs: add role and tools to getting started cli experience

psschwei:
Signed-off-by: Paul S. Schweigert <paul@paulschweigert.com>

Closes: #148

### Description

We have found that bees perform best when:
* they use fewer number of tools, and
* they are provided explicit instructions/role

Here we updated the getting started section of the README to incorporate those recommendations into the simple example. We update the system prompt to provide explicit instructions, and we remove the search tool as it isn't needed for weather-related queries. It also adds some tips related to these two points to the appropriate section.

Additionally, it adds a yarn helper method to run the simple example: `yarn run quickstart`

### Checklist

- [x] Tests are included 

no tests included as this is a documentation change

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ yarn add bee-agent-framework
 import { BeeAgent } from "bee-agent-framework/agents/bee/agent";
 import { OllamaChatLLM } from "bee-agent-framework/adapters/ollama/chat";
 import { TokenMemory } from "bee-agent-framework/memory/tokenMemory";
-import { DuckDuckGoSearchTool } from "bee-agent-framework/tools/search/duckDuckGoSearch";
+import { BeeSystemPrompt } from "bee-agent-framework/agents/bee/prompts";
 import { OpenMeteoTool } from "bee-agent-framework/tools/weather/openMeteo";
 
 const llm = new OllamaChatLLM(); // default is llama3.1 (8B), it is recommended to use 70B model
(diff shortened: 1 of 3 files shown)


psschwei:
FYI @mmurad2 - if this wasn't what you were envisioning let me know, happy to iterate on this

psschwei:
this is a new one :smile: 

image

image

README.md
```diff
@@ -74,9 +83,15 @@ const response = await agent
...
+> [!TIP]
+>
+> Providing a bee with specific `instructions` on what their role should be will improve performance and is recommended.
+
+> [!TIP]
+>
+> Bees work best when using a smaller number of `tools`, so it is recommended to only enable the minimum required tools.
 
-➡️ you can run this example after local installation, using the command `yarn start examples/agents/simple.ts`
+➡️ you can run this example after local installation, using the command `yarn run quickstart`
```
Tomas2D, line 94: `run` is not needed

README.md
```diff
@@ -86,6 +101,8 @@ console.log(`Agent 🤖 : `, response.result.text);
 >
 > Documentation is available at https://i-am-bee.github.io/bee-agent-framework/
 
+➡️ See a more [advanced example](/examples/agents/bee.ts)
```
Tomas2D, line 104: Missing `.` at the end.

README.md
```diff
@@ -52,15 +52,24 @@ yarn add bee-agent-framework
...
+  tools: [new OpenMeteoTool()], // for more explore 'bee-agent-framework/tools'
+  templates: {
+    system: BeeSystemPrompt.fork((old) => ({
+      ...old,
+      defaults: {
+        instructions:
+          "You are a helpful assistant that uses tools to answer weather-related questions.",
+      },
+    })),
+  },
```
Tomas2D, line 72: This definitely should not be part of README. Keep the example as is.

Tomas2D (changes requested):
I think the issue was about adding `role` and `tools` to the CLI.
So, when someone starts the script, it first asks for tools and a role, and then the chat begins.

psschwei:
> I think the issue was about adding `role` and `tools` to the CLI.
> So, when someone starts the script, it first asks for tools and a role, and then the chat begins.

Ok, I think I misunderstood the ask here.

Do we currently have a way to create an agent from the command line? I believe all the examples I saw here are running an existing source code file (e.g. `yarn start examples/agents/bee.ts`) which already have tools and instructions predefined. So I'd assume we'd need to edit the source code files to add the tools / role prior to running the file? Or is there another approach that I'm missing?

Would there be a way to use the API here? That seems like the ideal way to add this kind of functionality, but I'm assuming that would need the whole stack and not just the framework, so might not be an option here...

Tomas2D:
Why would you use an API in the framework? API is built on top of the framework.
Just ask the user to specify a role, and if one is not provided then fallback to the default. 

Maybe consult details in the related issue.

Issue #219, opened 2024-12-03
BeeAgent with granite LLM uses BeeAssistantPrompt vs. GraniteBeeAssistantPrompt


michael-desmond:
**Describe the bug**
I was noticing that BeeAgent when configured with a granite llm uses the standard BeeAssistantPrompt which conflicts with the granite system prompt and causes the llm to perform poorly. The offending code is here: 

https://github.com/i-am-bee/bee-agent-framework/blob/fd942ae0116893b9452530c305e0053bb83f5578/src/agents/bee/agent.ts#L133

A fairly simple fix is to conditionally choose a default AssistantPrompt before rendering i.e. 

```
const template = (this.input.llm.modelId.includes("granite"))?GraniteBeeAssistantPrompt: BeeAssistantPrompt;
```

But this is not the most elegant solution imo because it introduces more ad hoc checks into the agent. 

Ideally we could move the rendering of the assistant message into the runner hierarchy. The renders method seems to be the most appropriate spot. 

https://github.com/i-am-bee/bee-agent-framework/blob/fd942ae0116893b9452530c305e0053bb83f5578/src/agents/bee/runners/default/runner.ts#L256C17-L256C28

However the renders method is introduced on the DefaultRunner as opposed to the BaseRunner which is the type maintained by the BeeAgent. 

**To Reproduce**

1. Create a BeeAgent with granite and breakpoint line 130.
3. The BeeAssistantPrompt is applied when rendering the assistant prompt.

**Expected behavior**
The GraniteBeeAssistantPrompt should be applied when rendering the assistant prompt.

**Set-up:**

- Bee version: 0.0.47
- Model provider: Ollama

michael-desmond:
@Tomas2D Can u give an opinion on this one when you get a chance.

Tomas2D:
Good catch! 

Fixed in #223

Tomas2D:
Released in v0.0.48

PR #223, opened 2024-12-04
feat(agents): use prompt templates from a concrete runner

Tomas2D:
- Delegate template handling to a runner

Closes: #219

### DIFF
--- a/src/agents/bee/runners/granite/runner.ts
+++ b/src/agents/bee/runners/granite/runner.ts
@@ -18,7 +18,12 @@ import type { AnyTool } from "@/tools/base.js";
 import { isEmpty } from "remeda";
 import { DefaultRunner } from "@/agents/bee/runners/default/runner.js";
 import { BaseMemory } from "@/memory/base.js";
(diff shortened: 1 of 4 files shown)


src/agents/bee/runners/granite/runner.ts
```diff
@@ -83,6 +88,15 @@ export class GraniteRunner extends DefaultRunner {
     return memory;
   }
 
+  get templates(): BeeAgentTemplates {
```
michael-desmond, line 91: Now that we have this templates method, I wonder if we should be using this everywhere as opposed to using `this.input.templates`. If we use this templates method then we can removed the granite runner constructor code that duplicates this code.

https://github.com/i-am-bee/bee-agent-framework/blob/be3bb5c6d1a008cc205276b7ad0e6b759ae3ec97/src/agents/bee/runners/granite/runner.ts#L46C1-L48C84

Tomas2D: Yes, the template override is not necessary anymore. Removed.

michael-desmond (approved):

PR #235, opened 2024-12-09
chore(adapters): add llama3.3 template

jezekra1:


src/adapters/shared/llmChatTemplates.ts
```diff
@@ -96,6 +96,8 @@ const llama31: LLMChatTemplate = {
   },
 };
 
+const llama33: LLMChatTemplate = llama31;
```
Tomas2D, line 99: At least copy the object.

Tomas2D (approved):

PR #239, opened 2024-12-10
feat(agents): update streamlit agent prompt to write async code

jezekra1:


JanPokorny (approved):
There was a discrepancy between the prototype prompt and the framework prompt, I fixed it (hope you don't mind)

PR #244, opened 2024-12-11
fix(tools): verify that the end date is later than the start date

michael-desmond:

If the llm uses the openMeteo tool with a startDate in the future, but does not specify an endDate then the endDate is set to now which is prior to the start date. This causes the the service to throw an error. 

### Description

Check of the end data is prior to the start date, if so reset the end date. 

### Checklist

- [ ] Tests are included 

### DIFF
--- a/src/tools/weather/openMeteo.ts
+++ b/src/tools/weather/openMeteo.ts
@@ -179,7 +179,7 @@ export class OpenMeteoTool extends Tool<OpenMeteoToolOutput, ToolOptions, ToolRu
           const end = new Date(endDateStr);
           if (end < start) {
             throw new ToolInputValidationError(
-              `The 'end_date' (${endDateStr}) must be occur on or after the 'start_date' (${startDateStr}).`,
+              `The 'end_date' (${endDateStr}) has to occur on or after the 'start_date' (${startDateStr}).`,
             );
           }
           return { start: start, end: end };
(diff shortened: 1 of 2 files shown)


src/tools/weather/openMeteo.ts
```diff
@@ -169,10 +169,14 @@ export class OpenMeteoTool extends Tool<OpenMeteoToolOutput, ToolOptions, ToolRu
...
         : new Date(Date.UTC(now.getFullYear(), now.getMonth(), now.getDate()));
-      const end = endDate
+      let end = endDate
         ? new Date(endDate)
         : new Date(Date.UTC(now.getFullYear(), now.getMonth(), now.getDate()));
 
+      if (end < start) {
+        end = new Date(start);
+      }
+
```
Tomas2D, line 179: Good catch; however, we should rather throw an error and let the LLM regenerate.

```ts
throw new ToolInputValidationError(`The value of the 'end_date' property (${endDate}) must be greater than the value of 'start_date' (${startDate}).`);
```

What do you think?

Tomas2D (changes requested):

src/tools/weather/openMeteo.ts
michael-desmond: - If the llm provided an end date before the start date, then we should throw an error. 
- if the llm only provides a future start date, and we set an end data to satisfy the service which happens to be prior to the start date, then we should make sure that does not happen.

Setting the end date to start date + N will cause the service to provide weather data for the N days between start and end date which is probably not ideal. Ill fix the code to handle the 2 cases above.

Issue #238, opened 2024-12-09
Add support of Model Context Protocol (MCP)


ajshedivy:
Original Issue: https://github.com/i-am-bee/bee-stack/issues/59

MCP was recently open-sourced by Anthropic as a standard way for LLM applications to interact with external datasources. 
- [MCP docs](https://modelcontextprotocol.io/introduction)
- [Github](https://github.com/modelcontextprotocol/)

MCP offers additional compatibility for developing Agents and LLM tools (like database connectors)

**Describe the solution you'd like**
Allow Bee agents to integrate with external Tools via MCP. 

**Describe alternatives you've considered**
Bee already has great ways to interact with external tools, I built a sample agent that can interact with Db2 for i: https://github.com/ajshedivy/bee-agent-ibmi

In Bee stack I can simulate MCP via REST API. My workaround at the moment is to create a simple webserver on IBM i that can talk to Db2 that accepts requests like:

```bash
curl -X POST http://HOST:PORT/api/execute-sql \
     -H "Content-Type: application/json" \
     -d '{"sql": "SELECT * FROM sample.employee"}'
```
This will return the result set from the given SQL. 
 
I can then create a "proxy tool" for Bee that sends requests to this endpoint:

```python
import requests

def send_sql_query(sql_query) -> dict:
    """
    Sends a SQL statement to the Flask server and returns the result.

    Args:
        sql_query (str): The SQL statement to execute.

    Returns:
        dict: The server's response as a dictionary.
    """
    url = "http://HOST:PORT/api/execute-sql"  # Server endpoint
    headers = {"Content-Type": "application/json"}
    payload = {"sql": sql_query}

    try:
        response = requests.post(url, json=payload, headers=headers)
        response.raise_for_status() 
        return response.json()     
    except requests.exceptions.RequestException as e:
        return {"error": str(e)}

```

This works decent:
![image](https://github.com/user-attachments/assets/654d3781-9def-4432-b3fa-f13d439d353f)

**Additional context**
In summary, MCP gives developers the ability to maintain a single, standardized external tool that would allow for additional flexibility when it comes to creating Agents and LLM apps. This would be a great addition to the Bee framework 😁

pilartomas:
Hi @ajshedivy thanks for the issue, we're looking into it. 

[...]

PR #265, opened 2025-01-02
feat(tools): add Model Context Protocol tool

pilartomas:

Closes: #238 

### Description

This PR introduces support for [MCP Tool](https://modelcontextprotocol.io/docs/concepts/tools).

### Checklist

- [ ] I have read the [contributor guide](https://github.com/i-am-bee/bee-agent-framework/blob/main/CONTRIBUTING.md)
- [x] I have [signed off](https://github.com/i-am-bee/bee-agent-framework/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-dco) on my commit
- [x] Tests are included 

### DIFF
--- a/src/internals/helpers/paginate.ts
+++ b/src/internals/helpers/paginate.ts
@@ -40,3 +40,35 @@ export async function paginate<T>(input: PaginateInput<T>): Promise<T[]> {
 
   return acc;
 }
+
+export interface PaginateWithCursorInput<T, C> {
+  size: number;
+  handler: (data: {
+    cursor: C | undefined;
+    limit: number;
+  }) => Promise<{ data: T[]; done: true } | { data: T[]; done: false; nextCursor: C }>;
+}
+
+export async function paginateWithCursor<T, C>(input: PaginateWithCursorInput<T, C>): Promise<T[]> {
+  const acc: T[] = [];
+  let cursor: C | undefined;
+  while (acc.length < input.size) {
+    const result = await input.handler({
+      cursor,
+      limit: input.size - acc.length,
+    });
+    acc.push(...result.data);
+
+    if (result.done || result.data.length === 0) {
+      break;
+    } else {
+      cursor = result.nextCursor;
+    }
+  }
+
+  if (acc.length > input.size) {
+    acc.length = input.size;
+  }
+
+  return acc;
+}
--- a/src/internals/helpers/paginate.ts
+++ b/src/internals/helpers/paginate.ts
@@ -14,56 +14,25 @@
  * limitations under the License.
  */
 
-export interface PaginateInput<T> {
+export interface PaginateInput<T, C> {
   size: number;
-  handler: (data: { offset: number; limit: number }) => Promise<{ data: T[]; done: boolean }>;
+  handler: (data: { cursor?: C; limit: number }) => Promise<{ data: T[]; nextCursor?: C }>;
 }
 
-export async function paginate<T>(input: PaginateInput<T>): Promise<T[]> {
+export async function paginate<T, C = number>(input: PaginateInput<T, C>): Promise<T[]> {
   const acc: T[] = [];
-
+  let cursor: C | undefined = undefined;
   while (acc.length < input.size) {
-    const { data, done } = await input.handler({
-      offset: acc.length,
+    const { data, nextCursor } = await input.handler({
+      cursor,
       limit: input.size - acc.length,
     });
     acc.push(...data);
 
-    if (done || data.length === 0) {
+    if (nextCursor === undefined) {
       break;
     }
-  }
-
-  if (acc.length > input.size) {
-    acc.length = input.size;
-  }
-
-  return acc;
-}
-
-export interface PaginateWithCursorInput<T, C> {
-  size: number;
-  handler: (data: {
-    cursor: C | undefined;
-    limit: number;
-  }) => Promise<{ data: T[]; done: true } | { data: T[]; done: false; nextCursor: C }>;
-}
-
-export async function paginateWithCursor<T, C>(input: PaginateWithCursorInput<T, C>): Promise<T[]> {
-  const acc: T[] = [];
-  let cursor: C | undefined;
-  while (acc.length < input.size) {
-    const result = await input.handler({
-      cursor,
-      limit: input.size - acc.length,
-    });
-    acc.push(...result.data);
-
-    if (result.done || result.data.length === 0) {
-      break;
-    } else {
-      cursor = result.nextCursor;
-    }
+    cursor = nextCursor;
   }
 
   if (acc.length > input.size) {
--- a/src/internals/helpers/paginate.ts
+++ b/src/internals/helpers/paginate.ts
@@ -29,7 +29,7 @@ export async function paginate<T, C = number>(input: PaginateInput<T, C>): Promi
     });
     acc.push(...data);
 
-    if (nextCursor === undefined) {
+    if (nextCursor === undefined || data.length === 0) {
       break;
     }
     cursor = nextCursor;
--- a/docs/tools.md
+++ b/docs/tools.md
@@ -10,24 +10,24 @@ These tools extend the agent's abilities, allowing it to interact with external 
 
 ## Built-in tools
 
-| Name                                                                      | Description                                                                                                     |
-| ------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------- |
-| `PythonTool`                                                              | Run arbitrary Python code in the remote environment.                                                            |
-| `WikipediaTool`                                                           | Search for data on Wikipedia.                                                                                   |
-| `GoogleSearchTool`                                                        | Search for data on Google using Custom Search Engine.                                                           |
-| `DuckDuckGoTool`                                                          | Search for data on DuckDuckGo.                                                                                  |
-| [`SQLTool`](./sql-tool.md)                                                | Execute SQL queries against relational databases.                                                               |
-| `ElasticSearchTool`                                                       | Perform search or aggregation queries against an ElasticSearch database.                                        |
-| `CustomTool`                                                              | Run your own Python function in the remote environment.                                                         |
-| `LLMTool`                                                                 | Use an LLM to process input data.                                                                               |
-| `DynamicTool`                                                             | Construct to create dynamic tools.                                                                              |
-| `ArXivTool`                                                               | Retrieve research articles published on arXiv.                                                                  |
-| `WebCrawlerTool`                                                          | Retrieve content of an arbitrary website.                                                                       |
-| `OpenMeteoTool`                                                           | Retrieve current, previous, or upcoming weather for a given destination.                                        |
-| `MilvusDatabaseTool`                                                      | Perform retrieval queries (search, insert, delete, manage collections) against a MilvusDatabaseTool database.   |
-| `OpenAPITool`                                                             | Send requests to and receive responses from API server.                                                         |
-| `MCPTool`                                                                 | Discover and use tools exposed by arbitrary [MCP Server](https://modelcontextprotocol.io/examples) environment. |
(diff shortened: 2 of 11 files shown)


src/tools/mcp/mcpResource.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  async inputSchema() {
+    const resources = await this.listPaginatedResources().catch(() => []);
+    return z.object({
+      uri: z
+        .string()
+        .describe(
+          `URI of the resource to read, ${resources.length > 0 ? `available resources are:\n\n${resources.map(({ uri, name, description }) => JSON.stringify({ uri, name, description })).join("\n\n")}` : "no resources available at the moment"}.`,
+        ),
+    });
```
pilartomas, line 59: Suggestions are welcome

src/tools/mcp/mcpResource.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  public readonly resourceLimit: number;
+
+  constructor({ resourceLimit = 50, ...options }: MCPResourceToolInput) {
+    super(options);
+    this.resourceLimit = resourceLimit;
+  }
+
+  async inputSchema() {
+    const resources = await this.listPaginatedResources().catch(() => []);
```
pilartomas, line 45: Resource template support can be added later

src/tools/mcp/mcpResource.ts
```diff
@@ -0,0 +1,75 @@
...
+    });
+
+  public readonly resourceLimit: number;
+
+  constructor({ resourceLimit = Infinity, ...options }: MCPResourceToolInput) {
+    super(options);
+    this.resourceLimit = resourceLimit;
+  }
+
+  async inputSchema() {
```
jezekra1, line 44: We should probably cache this for some time, I can imagine `inputSchema` being called often while resources are changing less often

src/tools/mcp/mcpResource.ts
```diff
@@ -0,0 +1,75 @@
...
+    const resources = await this.listResources().catch((err) => {
+      if (err instanceof McpError && err.code === ErrorCode.MethodNotFound) {
+        return [];
+      }
+      throw err;
+    });
```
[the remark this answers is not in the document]
jezekra1: If the method for listing resources does not exist, the agent will need to know the uri of the resource exactly which I think will be rarely correct. 

I'd just say that the servers not implementing this method are not supported for now or they will require a new specialized implementation.

Ad limit - idea:
- wouldn't it be better to have a separate tool for listing resources? Pagination could be implemented as input parameter of the tool. Then the ResourceRead tool would have "you must first list resources" in the prompt

src/tools/mcp/base.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export abstract class MCPTool<T> extends Tool<MCPToolOutput<T>> {
+  public readonly client: MCPClient;
+
+  constructor({ client, ...options }: MCPToolInput) {
+    super(options);
+    this.client = client;
+  }
+
+  protected async paginateUntilLimit<T>(
```
Tomas2D, line 35: There is such utility function already. It is called `paginate` in `src/internals/helpers/paginate.ts`

src/tools/mcp/mcpResource.ts
```diff
@@ -0,0 +1,75 @@
...
+        .describe(
+          `URI of the resource to read, ${resources.length > 0 ? `available resources are:\n\n${resources.map(({ uri, name, description }) => JSON.stringify({ uri, name, description })).join("\n\n")}` : "no resources available at the moment"}.`,
+        ),
+    });
+  }
+
+  protected async _run({ uri }: ToolInput<this>, { signal }: { signal?: AbortSignal }) {
+    const result = await this.client.readResource({ uri }, { signal });
+    return new MCPToolOutput(result);
+  }
```
Tomas2D, line 63:
```ts
protected async _run({ uri }: ToolInput<this>, _options: BaseToolRunOptions, run: GetRunContext<typeof this>) {
    const result = await this.client.readResource({ uri }, { signal: run.signal });
    return new MCPToolOutput(result);
}
```

src/tools/mcp/mcpResource.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  public async listResources({ signal }: { signal?: AbortSignal } = {}) {
```
Tomas2D, line 65: This could be inlined to `_run`

Tomas2D (changes requested):

src/tools/mcp/mcpResource.ts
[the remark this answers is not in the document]
JanPokorny: Can't we let the developer hardcode the list of resources when instantiating the tool, as an alternative to listing them?

A separate tool for listing resources doesn't sound right to me, we want to be able to take advantage of constrained decoding and validation through the input schema, and also having more tools degrades performance.

src/tools/mcp/base.ts
pilartomas: ok, I'll add an overload for cursor based pagination.

src/tools/mcp/mcpResource.ts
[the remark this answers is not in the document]
pilartomas: > If the method for listing resources does not exist, the agent will need to know the uri of the resource exactly which I think will be rarely correct.
I'd just say that the servers not implementing this method are not supported for now or they will require a new specialized implementation.

Agreed, I'll remove the catch.

> wouldn't it be better to have a separate tool for listing resources? Pagination could be implemented as input parameter of the tool. Then the ResourceRead tool would have "you must first list resources" in the prompt

I think that would be too heavy on the agent, also the user would have to make sure to include both tools which is quite error prone (but we could solve it with a joint factory somehow).

> Can't we let the developer hardcode the list of resources when instantiating the tool, as an alternative to listing them?

We could but that would just put the problem on the user's back. The user would have to use the client, make a list a and inject the values into the constructor. It would also never update for that instance. I think we should solve it on our side. It would also sort of clash with another MCP concept from his perspective - [roots](https://modelcontextprotocol.io/docs/concepts/roots)

src/tools/mcp/mcpResource.ts
pilartomas: Assuming you mean `inputSchema` as this is used there, will do 👍

src/tools/mcp/mcpResource.ts
```diff
@@ -0,0 +1,78 @@
...
+import { GetRunContext } from "@/context.js";
+import { paginateWithCursor } from "@/internals/helpers/paginate.js";
+import { Cache } from "@/cache/decoratorCache.js";
+
+export interface MCPResourceToolInput extends MCPToolInput {
+  resourceLimit?: number;
+}
+
+export class MCPResourceTool extends MCPTool<ReadResourceResult> {
+  name = "Resource";
```
Tomas2D, line 31: `MCPResource`

src/tools/mcp/mcpResource.ts
```diff
@@ -0,0 +1,78 @@
...
+        return nextCursor
+          ? ({
+              data: resources,
+              done: false,
+              nextCursor,
+            } as const)
+          : ({
+              data: resources,
+              done: true,
+            } as const);
```
Tomas2D, line 62: Instead of passing `done` it would be much easier to pass falsy value for `nextCursor`.

src/tools/mcp/mcpResource.ts
Tomas2D: What about using `never` or `literal`?

- https://github.com/i-am-bee/bee-api/blob/c1e3c6dc86cab913c3a246c4cfafbe2456030355/src/runs/execution/tools/read-file-tool.ts#L51

src/tools/mcp/mcpResource.ts
```diff
@@ -0,0 +1,78 @@
...
+    return z.object({
+      uri: z
+        .string()
+        .describe(
+          `URI of the resource to read, ${resources.length > 0 ? `available resources are:\n\n${resources.map(({ uri, name, description }) => JSON.stringify({ uri, name, description })).join("\n\n")}` : "no resources available at the moment"}.`,
+        ),
+    });
+  }
+
+  protected async _run({ uri }: ToolInput<this>, run: GetRunContext<typeof this>) {
```
Tomas2D, line 74: Incorrect usage, `run` is a third parameter

Tomas2D: it should be 

```
protected async _run({ uri }: ToolInput<this>, _options: BaseToolRunOptions, run: GetRunContext<typeof this>) {
```

src/internals/helpers/paginate.ts
```diff
@@ -40,3 +40,35 @@ export async function paginate<T>(input: PaginateInput<T>): Promise<T[]> {
 
   return acc;
 }
+
+export interface PaginateWithCursorInput<T, C> {
+  size: number;
+  handler: (data: {
+    cursor: C | undefined;
```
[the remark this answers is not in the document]
pilartomas: The handler has the cursor defined on both input and output, this helps to tie the types of those variables.

src/internals/helpers/paginate.ts
```diff
@@ -40,3 +40,35 @@ export async function paginate<T>(input: PaginateInput<T>): Promise<T[]> {
...
+      cursor = result.nextCursor;
+    }
+  }
+
+  if (acc.length > input.size) {
+    acc.length = input.size;
+  }
+
+  return acc;
+}
```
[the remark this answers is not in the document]
pilartomas: Tried that initially but it was too cumbersome to use. What interface would you propose exactly?

src/tools/mcp/mcpResource.ts
pilartomas: Kept it in order to keep it close to the existing pagination but yeah technically missing cursor results in pagination being done.

pilartomas:
After further exploration, the resource capability is not meant to be model-controlled, I'll rework the tool to use tools capability which will also allow it to use actions.

src/tools/mcp/mcpResource.ts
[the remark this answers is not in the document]
pilartomas: After further dive into documentation, I've decided to switch the tool to target `tool` capability which is more native to models. This makes the following suggestion very much valid 👍 

> Can't we let the developer hardcode the list of resources when instantiating the tool, as an alternative to listing them?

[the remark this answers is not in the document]
ajshedivy: @pilartomas Catching up on this. I agree `tool` capabilities should be first priority. If need be, I think we can use `client.getServerCapabilities` to check what methods are provided by the server. 
https://github.com/search?q=repo%3Amodelcontextprotocol%2Ftypescript-sdk%20getServerCapabilities&type=code

src/internals/helpers/paginate.ts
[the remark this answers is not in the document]
Tomas2D: Just this `{ data: T[], nextCursor?: C }`

src/internals/helpers/paginate.ts
```diff
@@ -14,24 +14,25 @@
...
-    const { data, done } = await input.handler({
-      offset: acc.length,
+    const { data, nextCursor } = await input.handler({
+      cursor,
       limit: input.size - acc.length,
     });
     acc.push(...data);
 
-    if (done || data.length === 0) {
+    if (nextCursor === undefined) {
```
pilartomas, line 32: Empty data check has been removed. In cursor-based pagination, existence of the cursor should be the only stop condition (from the producers perspective). Therefore, handler is responsible for this.

src/tools/mcp/mcpResource.ts
pilartomas: done

src/internals/helpers/paginate.ts
Tomas2D: I would keep it just in case..

Tomas2D (approved):
Great work guys!

Issue #144, opened 2024-11-05
tools(similarity): filter out low scoring chunks


matoushavlena:
The wikipedia tool might be returning chunks that have low similarity scores and therefore not relevant/useful.

Similarly to `minPageNameSimilarity`, we would like to introduce a threshold that would filter out chunks/documents from the similarity tool. The initial value could be 0.25, but some exploration might be needed to decide on the right threshold.

When no documents are returned, the underlying tools (Wikipedia in this case) should return a LLM friendly message, such as "No results were found. Try to reformat your query.". This message already exists for the Wikipedia tool when no relevant pages are returned. We need to keep it DRY.

PR #266, opened 2025-01-02
feat(tools): add minScore filter condition for SimilarityTool

Tomas2D:
Ref: #144

### DIFF
--- a/src/tools/similarity.ts
+++ b/src/tools/similarity.ts
@@ -100,7 +100,7 @@ export class SimilarityTool<TProviderOptions> extends Tool<
         documentIndex: idx,
         score,
       })),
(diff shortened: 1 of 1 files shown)


src/tools/similarity.ts
```diff
@@ -99,6 +100,7 @@ export class SimilarityTool<TProviderOptions> extends Tool<
         documentIndex: idx,
         score,
       })),
+      filter(({ score }) => score >= (options.minScore ?? 0)),
```
JanPokorny, line 103: Is the score always positive? Maybe negative infinity would be a safer sentinel here

Tomas2D: Good catch. Updated.

JanPokorny (approved):

PR #267, opened 2025-01-06
fix(tools): add proper wikipedia result types

jezekra1:
Ref: https://github.com/i-am-bee/bee-api/pull/143

### DIFF
--- a/src/internals/types.ts
+++ b/src/internals/types.ts
@@ -95,6 +95,8 @@ export type PromiseOrPlain<T, Decider extends AnyFn> =
 
 export type NoPromise<T> = T extends Promise<unknown> ? never : T;
 
+export type UnwrapPromise<T> = T extends Promise<infer U> ? U : T;
+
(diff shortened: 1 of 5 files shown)


src/internals/types.ts
```diff
@@ -95,6 +95,8 @@ export type PromiseOrPlain<T, Decider extends AnyFn> =
 
 export type NoPromise<T> = T extends Promise<unknown> ? never : T;
 
+export type UnwrapPromise<T> = T extends Promise<infer U> ? U : T;
```
JanPokorny, line 98: Can you use `Awaited<T>`? https://www.typescriptlang.org/docs/handbook/utility-types.html

It's not 100% the same thing (it also unwraps nested promises recursively) but it might be what you want to use nonetheless

jezekra1: Nice, thanks!

JanPokorny: You forgot to remove the utility class 😁

JanPokorny (approved):

PR #270, opened 2025-01-09
feat(agents): improve template overriding

jezekra1:


src/template.ts
```diff
@@ -55,9 +54,9 @@ type PromptTemplateConstructor<T extends ZodType, N> = N extends ZodType
     }
   : Omit<PromptTemplateInput<T>, "schema"> & { schema: T | SchemaObject };
 
-type Customizer<T extends ZodType, N> = (
-  config: Required<PromptTemplateInput<T>>,
-) => PromptTemplateConstructor<T, N>;
+type Customizer<T extends ZodType, N> =
+  | ((config: Required<PromptTemplateInput<T>>) => PromptTemplateConstructor<T, N>)
+  | ((config: Required<PromptTemplateInput<T>>) => void);
```
Tomas2D, line 59: ⚠️ This could cause serious bug. `void` essentially means I do not care about the return value but in your implementation you are having this `const newConfig = customizer?.(config) ?? config;`

```
template.fork(() => 42) // typed validation OK, runtime fails
```

Tomas2D: I would update the `fork` implementation to throw if the return type is non-nullish and not a config object.

Tomas2D:
Good job! 🚀 👍🏻

Issue #334, opened 2025-02-13
feat(tools): implement Wikipedia Tool


Tomas2D:
- mirror TS behavior (can be simplified as the TS one is complex)
- ideally use some helper library
- don't forget to use async processing (HTTP)

PR #344, opened 2025-02-20
feature: add initial wikipedia search tool

AngeloDanducci:

Closes: #334 

### Checklist

- [x] Tests are included 


python/beeai_framework/tools/search/wikipedia.py
```diff
@@ -13,16 +13,47 @@
...
+    pass
 
 
-class WikipediaTool(Tool):
+class WikipediaSearchTool(Tool[WikipediaSearchToolInput]):
     name = "Wikipedia"
-    description = "Search factual and historical information, including biography, history, politics, geography, society, culture, science, technology, people, animal species, mathematics, and other subjects."  # noqa: E501
+    description = "Search factual and historical information, including biography, history, politics, geography, society, culture, science, technology, people, animal species, mathematics, and other subjects."
+    input_schema = WikipediaSearchToolInput
+    wiki_wiki = wikipediaapi.Wikipedia(user_agent="beeai-framework https://github.com/i-am-bee/beeai-framework", language='en')
```
Tomas2D, line 41: - What about just `client`?
- allow people to pass their options via the constructor (`__init__`)

Tomas2D (changes requested):
Good work so far! Just a few improvements, and we can merge it.

Tomas2D:
Link issue via `Ref: #...` in the issue description

AngeloDanducci:
Hmm didn't initially run into the git commit message error since it was failing on linting I will go back and fix that.

I think it's worth discussing how strict we want to be with commit messages - which probably also includes a discussion on if we want to define poe as the standard way to handle git commands.

Tomas2D:
- `poe` is script manager to simplify working with messages
- if you use `poe commit` it will help you to get it right
- maybe we don't need to check all commits, but just the latest

AngeloDanducci:
If we want the commit history to be uniform I think it makes sense to check them all.

Maybe we can fail the lint of the commit message before anything else so there's less of a chance of having to fix multiple commit messages in case people are committing from IDE / not using the poe tool.

python/poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 2.1.1 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.0.1 and should not be changed by hand.
```
planetf1, line 1: General question - what is our strategy on `poetry.lock`. 
pro - it pins the resolution of dependencies, and reduces change of varying resolution when installed by end user
cons - it can be affected by the last dev who updates (for example here the version used differs - minor), unless done in ci
...?

python/pyproject.toml
```diff
@@ -59,7 +60,6 @@ major_version_zero = true
 bump_message = "chore: python release $new_version"
 
 [tool.pytest.ini_options]
-addopts = "-v --cov --cov-report html"
```
planetf1, line 62: We're loosing the code coverage options?

Tomas2D:
> Maybe we can fail the lint of the commit message before anything else so there's less of a chance of having to fix multiple commit messages in case people are committing from IDE / not using the power tool.

If you had installed the git hooks (`.githooks/install.sh`), the GIT would have prevented you from committing.

python/poetry.lock
Tomas2D: I would pin it definitelly.

python/beeai_framework/tools/search/wikipedia.py
```diff
@@ -13,16 +13,54 @@
...
+    client = wikipediaapi.Wikipedia(
+        user_agent="beeai-framework https://github.com/i-am-bee/beeai-framework", language="en"
+    )
 
-    def input_schema(self) -> str:
-        # TODO: remove hard code
-        return '{"type":"object","properties":{"query":{"type":"string","format":"date","description":"Name of the wikipedia page, for example \'New York\'"}}}'  # noqa: E501
+    def __init__(self, full_text: bool = False) -> None:
+        super().__init__()
+        self.full_text = full_text
```
Tomas2D, line 50: Use `BaseModel` object instead (`WikipediaInput` / `WikipediaOptions`) to prevent further breaking changes.
Add more options that the client provides.

python/beeai_framework/tools/search/wikipedia.py
```diff
@@ -13,16 +13,54 @@
...
+                search_results: list[WikipediaSearchToolResult] = [
+                    WikipediaSearchToolResult(
+                        title=input.query or "", description=description_output or "", url=page_py.fullurl or ""
+                    )
+                ]
+                return WikipediaSearchToolOutput(search_results)
+            else:
+                raise Exception(f"No Wikipedia page matched the search term: {input.query}.")
+        except Exception as e:
+            raise ToolError("Error performing Wikipedia search: ") from e
```
Tomas2D, line 66: Not needed as exceptions are catched in the `run` method on your behalf

python/beeai_framework/tools/search/wikipedia.py
```diff
@@ -13,16 +13,54 @@
...
+
+class WikipediaSearchToolResult(SearchToolResult):
+    pass
+
+
+class WikipediaSearchToolOutput(SearchToolOutput):
+    pass
+
+
+class WikipediaSearchTool(Tool[WikipediaSearchToolInput]):
```
Tomas2D, line 38: I would just call it `WikipediaTool`

Tomas2D (changes requested):

Tomas2D:
Mention example in `docs/examples/README.md`

Tomas2D (approved):

Issue #331, opened 2025-02-17
LinePrefixParser state transitions + test cases


michael-desmond:
The line prefix parser implementation does not support the definition of valid start states or valid transitions. Need to improve the parser input specification to indicate start states (e.g. thought) and to indicate valid transitions (e.g thought -> final answer = ok). 

We also want to port over the test cases from the .ts implementation and make sure they pass.

PR #359, opened 2025-02-20
feat(parsers): add line prefix parser

Tomas2D:
Closes: #331

### DIFF
--- a/python/beeai_framework/agents/runners/default/runner.py
+++ b/python/beeai_framework/agents/runners/default/runner.py
@@ -14,7 +14,6 @@
 
 import json
 from collections.abc import Callable
-from typing import Any
 
 from beeai_framework.agents.runners.base import (
     BaseRunner,
@@ -39,11 +38,11 @@ from beeai_framework.agents.types import (
 )
 from beeai_framework.backend.chat import ChatModelInput, ChatModelOutput
 from beeai_framework.backend.message import SystemMessage, UserMessage
-from beeai_framework.emitter.emitter import Emitter, EventMeta
+from beeai_framework.emitter.emitter import EventMeta
 from beeai_framework.memory.base_memory import BaseMemory
 from beeai_framework.memory.token_memory import TokenMemory
 from beeai_framework.parsers.field import ParserField
-from beeai_framework.parsers.line_prefix import LinePrefixParser, LinePrefixParserNode
+from beeai_framework.parsers.line_prefix import LinePrefixParser, LinePrefixParserNode, LinePrefixParserUpdate
 from beeai_framework.tools import ToolError, ToolInputValidationError
 from beeai_framework.tools.tool import StringToolOutput, Tool, ToolOutput
 from beeai_framework.utils.strings import create_strenum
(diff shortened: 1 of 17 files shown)


python/beeai_framework/emitter/emitter.py
```diff
@@ -182,7 +185,10 @@ async def invoke(self, data: Any, event: EventMeta) -> None:
                 self.listeners.remove(listener)
 
             async def run(ln: Listener = listener) -> Any:
-                return await ln.callback(data, event)
+                if inspect.iscoroutinefunction(ln.callback):
+                    return await ln.callback(data, event)
+                else:
+                    return ln.callback(data, event)
```
Tomas2D, line 191: could be extracted to another PR
cc @ajbozarth

python/beeai_framework/emitter/types.py
```diff
@@ -30,8 +30,10 @@ class EmitterInput(BaseModel):
...
 
 
-class EmitterOptions(BaseModel, frozen=True):
+class EmitterOptions(BaseModel):
     is_blocking: bool | None = None
     once: bool | None = None
     persistent: bool | None = None
     match_nested: bool | None = None
+
+    model_config = ConfigDict(frozen=True)
```
Tomas2D, line 39: could be extracted to another PR
cc @ajbozarth

ajbozarth: This was what I was trying to do, I can create a separate PR with this and some other small fixes if you don't want it blocked on this PR

python/beeai_framework/emitter/emitter.py
ajbozarth: I was about to argue against this because I thought `ln.callback` had to be async, but upon review it actually doesn't. This would also allow us to remove async from a bunch of the callback defs. I can also include this in a separate cleanup PR. I'll have one open by EOD for review before release tomorrow

ajbozarth:
I opened #361 which includes the bug fixes here so this PR doesn't block the release

michael-desmond:
I had a look through and I noticed that the default runner still has the old llm method which seems to be broken now. 

```
AttributeError: 'LinePrefixParser' object has no attribute 'feed'
```

https://github.com/i-am-bee/beeai-framework/blob/385b4e56342c7c2dc23f8862fe253d09bc18aca4/python/beeai_framework/agents/runners/default/runner.py#L92

python/beeai_framework/agents/runners/default/runner.py
```diff
@@ -58,15 +60,34 @@ def default_templates(self) -> BeeAgentTemplates:
...
+                "tool_output": LinePrefixParserNode(
+                    prefix="Function Input: ", field=ParserField.from_type(str), is_end=True, next=["final_answer"]
+                ),
+                "final_answer": LinePrefixParserNode(
+                    prefix="Final Answer: ", field=ParserField.from_type(str), is_end=True, is_start=True
+                ),
+            }
+        )
 
     async def llm(self, input: BeeRunnerLLMInput) -> BeeAgentRunIteration:
```
michael-desmond, line 92: I believe that the llm method should be updated to use the new parser.

python/beeai_framework/agents/runners/granite/runner.py
```diff
@@ -25,19 +25,39 @@
...
+                "tool_name": LinePrefixParserNode(
+                    prefix="Tool Name: ",
+                    field=ParserField.from_type(tool_names, lambda v: v.trim()),
+                    next=["tool_input"],
+                ),  # validate enum
+                "tool_input": LinePrefixParserNode(
+                    prefix="Tool Input: ", field=ParserField.from_type(dict), next=["tool_output"]
+                ),
+                "tool_output": LinePrefixParserNode(
+                    prefix="Function Input: ", field=ParserField.from_type(str), is_end=True, next=["final_answer"]
```
michael-desmond, line 54: "Tool Output: " ?

python/beeai_framework/agents/runners/default/runner.py
```diff
@@ -58,15 +60,34 @@ def default_templates(self) -> BeeAgentTemplates:
...
+                "tool_input": LinePrefixParserNode(
+                    prefix="Function Input: ",
+                    field=ParserField.from_type(dict),
+                    next=["tool_output"],
+                ),
+                "tool_output": LinePrefixParserNode(
+                    prefix="Function Input: ", field=ParserField.from_type(str), is_end=True, next=["final_answer"]
+                ),
+                "final_answer": LinePrefixParserNode(
+                    prefix="Final Answer: ", field=ParserField.from_type(str), is_end=True, is_start=True
```
michael-desmond, line 87: Final Answer is not a start node afaik

python/beeai_framework/agents/runners/granite/runner.py
```diff
@@ -25,19 +25,39 @@
...
+                    is_start=True,
+                    next=["tool_name", "final_answer"],
+                ),
+                "tool_name": LinePrefixParserNode(
+                    prefix="Tool Name: ",
+                    field=ParserField.from_type(tool_names, lambda v: v.trim()),
+                    next=["tool_input"],
+                ),  # validate enum
+                "tool_input": LinePrefixParserNode(
+                    prefix="Tool Input: ", field=ParserField.from_type(dict), next=["tool_output"]
```
michael-desmond, line 51: For granite isnt the tool_input a terminal state?

python/beeai_framework/agents/runners/granite/runner.py
```diff
@@ -25,19 +25,39 @@
...
+                ),
+                "tool_output": LinePrefixParserNode(
+                    prefix="Function Input: ", field=ParserField.from_type(str), is_end=True, next=["final_answer"]
+                ),
+                "final_answer": LinePrefixParserNode(
+                    prefix="Final Answer: ", field=ParserField.from_type(str), is_end=True, is_start=True
```
michael-desmond, line 57: is_start should be false

michael-desmond (changes requested):

python/beeai_framework/agents/runners/default/runner.py
Tomas2D: It is because we are using fallback. If LLM replies with just the `Final Answer` we do the following fallback.

```python
def fallback(value: str) -> list[dict]:
    return [
        {"key": "thought", "value": "I now know the final answer."},
        {"key": "final_answer", "value": value},
    ]
```

python/beeai_framework/agents/runners/granite/runner.py
Tomas2D: not anymore, we are using `fallback` functionality so having just a `Final Answer` is okay.

python/beeai_framework/agents/runners/default/runner.py
Tomas2D: good catch, updated

michael-desmond (approved):

ajbozarth (approved):
This LGTM

IIUC the runner updates address bullet 1 in #362

Also I like the `to_template` util, is it something we should make more generically available for other places or is it specific to that use case?

PR #369, opened 2025-02-21
fix: apply suggestions from mypy

Tomas2D:


python/beeai_framework/context.py
```diff
@@ -45,56 +44,52 @@ class RunContextInput(BaseModel):
...
         self.tasks.append((fn, self.run_context))
         return self
 
     async def _run_tasks(self) -> R:
         for fn, param in self.tasks:
-            if inspect.iscoroutinefunction(fn):
-                await fn(param)
-            else:
-                fn(param)
+            await ensure_async(fn)(param)
```
ajbozarth, line 71: ty this is the cleaner solution I was searching for

python/beeai_framework/context.py
```diff
@@ -45,56 +44,52 @@ class RunContextInput(BaseModel):
...
         self.run_context.emitter.context = context
 
 
 class RunContext(RunInstance):
     storage: ContextVar[Self] = ContextVar("storage", default=None)
 
-    def __init__(
-        self, *, instance: RunInstance, context_input: RunContextInput, parent: GetRunContext | None = None
-    ) -> None:
+    def __init__(self, *, instance: RunInstance, context_input: RunContextInput, parent: Self | None = None) -> None:
```
ajbozarth, line 84: ty for finding the solution to get rid of `GetRunContext` it kept eluding me when I tried

python/beeai_framework/errors.py
```diff
@@ -59,114 +60,48 @@ def name(self) -> str:
         return type(self).__name__
 
     def has_fatal_error(self) -> bool:
-        """
-        Check if this error or any in the chain of exceptions under __cause__ is fatal (iterative).
-        """
-        current_exception = self  # Start with the current exception
+        current_exception: BaseException | None = self
```
ajbozarth, line 63: is there a reason we're removing the doc strings? I don't see a reason not to leave them if they are already written

ajbozarth (approved):
This LGTM, just a handful of TY comments and one nit about removing docstrings

python/beeai_framework/errors.py
Tomas2D: The method names seem self-descriptive to me. In addition, the implementation is short.

python/beeai_framework/context.py
Tomas2D: same story here, did my best to get rid of it

PR #379, opened 2025-02-24
docs: Corrections in top-level & python notebook readmes

planetf1:

Closes: n/a

### Description

- README.md
  - contribution guidelines pointed to typescript version. We have two different sets of docs for python and typescript. As an immediate fix, Added unique links to both. Longer term We could consider a common file at top level, and move common topics there before pointing to python/typescript specifics
  - Samples directory was also the typescript version. Followed same approach as above - two links
  - Link to multiagent ts example was pointing to readme. corrected to point to most similar ts example (weather workflow)
- python/examples/notebooks
  - fixed coalesced bullet points for installing ollama + model (and expanded text to look more consistent)
  
 Alternatively the lists could be collapsed to a simple statement. 

### Checklist

- [ ] Unit tests pass: `yarn test:unit`
- [ ] Tests are included 

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,7 @@ if __name__ == "__main__":
 
 _Source: [python/examples/workflows/multi_agents.py](/python/examples/workflows/multi_agents.py)_
 
-TypeScript version of this example can be found [here](/typescript/README.md).
+TypeScript version of this example can be found [here](/typescript/examples/workflows/multiAgents.ts).
 
 ### Running the example
 
--- a/README.md
+++ b/README.md
@@ -132,9 +132,7 @@ To run projects, use:
 python [project_name].py
 ```
 
-➡️ Explore more in our: 
-  - [Python examples library](/python/examples) 
-  - [TypeScript examples library](/typescript/examples)
+Explore more in our examples for [Python](/python/examples) and [TypeScript](/typescript/examples).
 
 ## Roadmap
 
--- a/README.md
+++ b/README.md
@@ -132,7 +132,9 @@ To run projects, use:
 python [project_name].py
 ```
 
-➡️ Explore more in our [examples library](/typescript/examples).
+➡️ Explore more in our: 
+  - [Python examples library](/python/examples) 
+  - [TypeScript examples library](/typescript/examples)
 
 ## Roadmap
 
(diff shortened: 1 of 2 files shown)


planetf1:
Noticed another error (not in this pR, additional issue). News item has an poor link
2025-02-19: Launched Python library alpha. See [getting started guide](https://github.com/i-am-bee/beeai-framework/blob/main/python/docs)

This points to the directory, but there's no guide. There is an `index.html`, and an overview, but these are more of a catalogue than a guide.

planetf1:
@jenna these are some minor updates I noticed. Ok with them? Or are they covered by your changes?

README.md
```diff
@@ -130,7 +130,9 @@ To run projects, use:
 python [project_name].py
 ```
 
-➡️ Explore more in our [examples library](/typescript/examples).
+➡️ Explore more in our: 
+  - [Python examples library](/python/examples) 
+  - [TypeScript examples library](/typescript/examples)
```
grahamwhiteuk, line 135: I wonder whether having Python and TypeScript called out all over our documentation in separate lists is going to make the docs a bit messy and tiresome to read?  I'm not sure whether it will so this is just an open comment for discussion.  Is the form we've got here better than a single sentence along the lines of

> Explore more in our [Python examples](/python/examples) and [TypeScript examples](/typescript/examples).

On reflection, I think the latter is more succinct and easier to read for me - what do others think?

jenna-winkler: I agree, we had discussed only sharing one language on this top level README (ideally Python when it’s in better shape) for this reason

planetf1: I certainly find it confusing at the moment.

One additional consideration is we need to be careful with text and section/page titles, as this affects the searchability - as a python developer I'd want to mostly see concepts and python examples, or at least clearly know which to click on.

On a previous project we were using mkdocs, and were able to use a tabbed layout for code examples, as well as expandable sections - so effectively python/typescript would be in the same place. 

Whichever approach we have needs to be consistent - if we go with both, that should apply for everything we talk about in python, and linked in the same way.

planetf1: As a FYI this PR is probably superceeded by @jenna-winkler - opened with some obvious errors when I thought the deadline was very close.

planetf1:
The minor corrections here still seem valid for the notebook (@grahamwhiteuk ) and links in readme (@jenna-winkler ) though on the latter there's still more to do to simplify the flow. I simplified the link to samples, though for now left contributing . review requested

Tomas2D (approved):

PR #404, opened 2025-02-26
feat: improve pre-commit hooks

Tomas2D:
- improve the portability of git hooks
- handle relative paths
- auto stage files


typescript/.yarn/install-state.gz
Tomas2D, line 1: This file should be ignored, but got there because of the monorepo migration

xjacka (approved):

Tomas2D:
cc @ajbozarth

PR #406, opened 2025-02-26
docs(tools): update tool docs and add missing examples

AngeloDanducci:

Contributes-to: https://github.com/i-am-bee/internal/issues/104

### Description

update tool docs and add missing examples

### Checklist

- [ ] Tests are included 

### DIFF
--- a/python/docs/tools.md
+++ b/python/docs/tools.md
@@ -44,7 +44,7 @@ if __name__ == "__main__":
 
 ```
 
-_Source: [python/examples/tools/base.py](python/examples/tools/base.py)_
+_Source: [/python/examples/tools/base.py](/python/examples/tools/base.py)_
 
 ### Advanced
 
@@ -71,7 +71,7 @@ if __name__ == "__main__":
 
 ```
 
-_Source: [python/examples/tools/advanced.py](python/examples/tools/advanced.py)_
+_Source: [/python/examples/tools/advanced.py](/python/examples/tools/advanced.py)_
 
 > [!TIP]
 >
@@ -105,7 +105,7 @@ if __name__ == "__main__":
 
 ```
 
-_Source: [python/examples/tools/agent.py](python/examples/tools/agent.py)_
+_Source: [/python/examples/tools/agent.py](/python/examples/tools/agent.py)_
 
 ### Usage with decorator
 
@@ -172,7 +172,7 @@ if __name__ == "__main__":
 
 ```
 
-_Source: [python/examples/tools/decorator.py](python/examples/tools/decorator.py)_
+_Source: [/python/examples/tools/decorator.py](/python/examples/tools/decorator.py)_
 
 ### Usage with duckduckgo
 
@@ -202,7 +202,7 @@ if __name__ == "__main__":
 
 ```
 
-_Source: [python/examples/tools/duckduckgo.py](python/examples/tools/duckduckgo.py)_
+_Source: [/python/examples/tools/duckduckgo.py](/python/examples/tools/duckduckgo.py)_
 
 ### Usage with openmeteo
 
@@ -232,7 +232,7 @@ if __name__ == "__main__":
 
 ```
 
-_Source: [python/examples/tools/openmeteo.py](python/examples/tools/openmeteo.py)_
+_Source: [/python/examples/tools/openmeteo.py](/python/examples/tools/openmeteo.py)_
 
 
 ### Usage with Wikipedia
@@ -260,13 +260,13 @@ if __name__ == "__main__":
 
 ```
 
-_Source: [python/examples/tools/wikipedia.py](python/examples/tools/wikipedia.py)_
+_Source: [/python/examples/tools/wikipedia.py](/python/examples/tools/wikipedia.py)_
 
 ## Writing a new tool
 
 To create a new tool it is recommended to implement the base `Tool` class.  
 
-- Initiate the [`Tool`](/beeai-framework/python/beeai_framework/tools/tool.py) by passing your own handler (function) with the `name`, `description` and `input schema`.
+- Initiate the [`Tool`](/python/beeai_framework/tools/tool.py) by passing your own handler (function) with the `name`, `description` and `input schema`.
 
 #### Basic
 
@@ -318,7 +318,7 @@ if __name__ == "__main__":
 
 ```
 
-_Source: [python/examples/tools/custom/base.py](python/examples/tools/custom/base.py)_
+_Source: [/python/examples/tools/custom/base.py](/python/examples/tools/custom/base.py)_
 
 > [!TIP]
 >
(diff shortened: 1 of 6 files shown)


AngeloDanducci:
Still need to add the equivalent of `openLibrary` from the typescript version but should be ready for lookover.

planetf1 (approved):

python/docs/tools.md
```diff
@@ -14,61 +12,105 @@ These tools extend the agent's abilities, allowing it to interact with external
...
-> Would you like to use a tool from LangChain? See the [example](/examples/tools/langchain.py).
+
+All examples can be found [here](/python/examples/tools).
 
 ## Usage
 
 ### Basic
 
+<!-- embedme examples/tools/base.py -->
+From [base.py](examples/tools/base.py):
```
Tomas2D, line 29: remove `From` as you already have `Source: `, update everywhere please

Tomas2D (changes requested):

python/examples/tools/base.py
```diff
@@ -0,0 +1,13 @@
+import asyncio
+
+from beeai_framework.tools.weather.openmeteo import OpenMeteoTool, OpenMeteoToolInput
+
+
+async def main() -> None:
+    tool = OpenMeteoTool()
+    result = tool.run(input=OpenMeteoToolInput(location_name="New York"))
```
ajbozarth, line 8: is there a reason you left out `start_date` and `end_date` inputs? They're in the TS implementation

python/examples/tools/agent.py
```diff
@@ -0,0 +1,20 @@
...
+from beeai_framework.agents.bee import BeeAgent
+from beeai_framework.agents.types import BeeInput, BeeRunInput
+from beeai_framework.backend.chat import ChatModel
+from beeai_framework.memory import UnconstrainedMemory
+from beeai_framework.tools.weather.openmeteo import OpenMeteoTool
+
+
+async def main() -> None:
+    llm = ChatModel.from_name("ollama:3.1")
+    agent = BeeAgent(BeeInput(llm=llm, tools=[OpenMeteoTool()], memory=UnconstrainedMemory()))
```
ajbozarth, line 12: I did noticed this is different from TS, but given how pointless the TS example is, I understand the choice. @jenna-winkler Is this difference and issue? this is what TS looks like: https://github.com/i-am-bee/beeai-framework/blob/main/typescript/examples/tools/agent.ts seems like it's a "docs only" example and not meant to run on it's own

python/examples/tools/advanced.py
```diff
@@ -0,0 +1,17 @@
+import asyncio
+
+from beeai_framework.tools.weather.openmeteo import OpenMeteoTool, OpenMeteoToolInput
+
+
+async def main() -> None:
+    tool = OpenMeteoTool()
```
ajbozarth, line 7: compared to TS you skipped the options input here and removed the following logging, is there a reason?

python/examples/tools/custom/base.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def _run(self, input: RiddleToolInput, _: Any | None = None) -> None:
+        index = input.riddle_number % (len(self.data))
+        riddle = self.data[index]
+        return riddle
+
+
+async def main() -> None:
+    tool = RiddleTool()
+    input = RiddleToolInput(riddle_number=1)
```
ajbozarth, line 36: This is an intriguing way of reimagining the TS implementation, but shouldn't this be a random value not `1`?

ajbozarth:
```suggestion
    input = RiddleToolInput(riddle_number=random.randint(1, len(RiddleTool.data)))
```
and add `import random` to the imports

python/examples/tools/agent.py
```diff
@@ -0,0 +1,20 @@
...
+
+from beeai_framework.agents.bee import BeeAgent
+from beeai_framework.agents.types import BeeInput, BeeRunInput
+from beeai_framework.backend.chat import ChatModel
+from beeai_framework.memory import UnconstrainedMemory
+from beeai_framework.tools.weather.openmeteo import OpenMeteoTool
+
+
+async def main() -> None:
+    llm = ChatModel.from_name("ollama:3.1")
```
ajbozarth, line 11: I'm not sure this is valid, the example errors for me

python/examples/tools/custom/openlibrary.py
```diff
@@ -0,0 +1,63 @@
...
+    bib_key: str
+
+
+class OpenLibraryTool(Tool[OpenLibraryToolInput]):
+    name = "OpenLibrary"
+    description = """Provides access to a library of books with information about book titles,
+        authors, contributors, publication dates, publisher and isbn."""
+    input_schema = OpenLibraryToolInput
+
+    def _run(self, input: OpenLibraryToolInput, _: Any | None = None) -> None:
```
ajbozarth, line 29:
```suggestion
    def _run(self, input: OpenLibraryToolInput, _: Any | None = None) -> OpenLibraryToolResult:
```

python/examples/tools/custom/openlibrary.py
```diff
@@ -0,0 +1,63 @@
...
+        input_vars = vars(input)
+        for val in input_vars:
+            if input_vars[val] is not None:
+                key = val
+                value = input_vars[val]
+                break
+        else:
+            raise ToolInputValidationError("All input values in OpenLibraryToolInput were empty.") from None
+
+        response = requests.get(
```
ajbozarth, line 41: TS had emitters before and after this call, they seem unused though, is that why you left them out?

ajbozarth (commented):
A handful of questions inline, but comparing this to TS I'm wondering if some of these examples should not have a main at all (like in TS)

python/examples/tools/base.py
AngeloDanducci: No strong reason other than this being a base example, added for next commit.

python/examples/tools/custom/openlibrary.py
AngeloDanducci: They seemed unused and the focus seems to be more on the tool than an emitter.

python/examples/tools/custom/openlibrary.py
AngeloDanducci: Good catch missed updating this as I changed to using `OpenLibraryToolResult.`

python/examples/tools/agent.py
AngeloDanducci: Yeah the TS example uses the ArXiv tool instead but we don't have that yet.  Given that this seems to be just an example of how you would set up an agent with an llm, memory, and tools I don't think it matters.

python/examples/tools/agent.py
AngeloDanducci: I'll double check this.

AngeloDanducci: Hmm not sure what changed if anything since writing, but it definitely errors out: `LinePrefixParserError: The generated output does not adhere to the schema.`
 
 Looks like the same issue when trying to run the `example/tools/openmeteo.py` example as well and also the `examples/agents/simple.py` example.

ajbozarth (approved):

Tomas2D (approved):

Issue #336, opened 2025-02-19
feat: examples parity


Tomas2D:

ajbozarth:
I believe we can move this out of the [Hackathon milestone](https://github.com/i-am-bee/beeai-framework/milestone/1) and I'll continue work in it into next week

jenna-winkler:
@ajbozarth please also prioritize /python/examples/workflows/nesting.py

jenna-winkler:
Flagging a todo item within this sample code /python/examples/workflows/web_agent.py

<img width="498" alt="Image" src="https://github.com/user-attachments/assets/f1e9f44b-ee4c-4674-9dfe-d7fdfb4fa941" />

jenna-winkler:
priority 1 items above ^

priority 2:
Missing memory examples
- /examples/memory/base.py
- /examples/memory/llmMemory.py
- /examples/memory/custom.py

jenna-winkler:
I created a table in the body of this issue to keep things organized @ajbozarth @AngeloDanducci

ajbozarth:
FYI I added section folding to the tables at the top for easier reading. The LOWER PRIORITY tables are now hidden by default and any table can be folded or opened for easier reading while we work on this

ajbozarth:
The description reflects the status once the following PRs are merged:
- #494 
- #498
- #481 
- #474

planetf1:
I didn't see it in the list, but there is a `version.py` just containing a TODO.

planetf1:
There were basic examples for ollama/watsonx/openai. Recently I've added groq, xai, amazon bedrock, google vertex, microsoft openai (draft PR) providers, so have made sure to include samples as part of that . 

Anthropic is next on the list which I'll likely start looking at tomorrow. 

That gets us to a baseline, but there's a lot more we can do

Did you have any specific thoughts about the provider examples? In one regard they're little more than a one-shot technical example of a few usage patterns with a specific provider/model.  I think we can do more in futures and have some ideas we could discuss

ajbozarth:
I like how much more robust the python backend provider examples are. The TS versions are trivial

Tomas2D:
What is the current state? @AngeloDanducci , @jenna-winkler

ajbozarth:
The current state is that all the examples that can be brought over from TS have been and those that haven't have the class or feature blocking them listed in the table. 

We can either close this issue as fixed or we can leave it open to track the remaining examples as the features blocking them are added

[...]

PR #438, opened 2025-03-01
fix(examples): better error handling and other fixes

ajbozarth:

Ref #336

### Description

Various improvements to existing examples:
- Add better and more consistent error handling, this also cleans up pytest e2e output
- set `options.execution.max_iterations` default to `math.inf` to match TS implementation 
- various "mypy" fixes (actually just PyCharm checker)
- conditionally run watsonx and openai examples in pytest
- fix pytest that was whacking the ollama env vars for future tests (like examples)

### Checklist

- [x] Tests are included 


ajbozarth:
If I do any more work on existing examples this weekend I'll add here, otherwise I'll open separate PRs for new examples.

ajbozarth:
Also @Tomas2D the e2e tests are still failing, but in a smaller number now. The following examples fail intermittently when run (even outside pytest): 

- `agents/simple.py`
- `tools/duckduckgo.py`
- `tools/openmeteo.py`
- `workflows/multi_agents.py`

The first three usually fail on ToolInput validation, but sometimes on the old `LinePrefixParserError('The generated output does not adhere to the schema.\nNothing valid has been parsed yet!')`

`multi_agents.py` sometimes also fails on Tool validation, but is generally still failing the same way as `test_multi_agents_workflow_creation`, which from a bit more testing seems to be at somewhat caused by the existence of multiple agents in the workflow.

Tomas2D (changes requested):
Can you move all try/except blocks within example body into the `__main__` block?
It would be easier to read and generally look better in the docs (in the future we might stip the `__main__` block entirely in the documentation)

ajbozarth:
> Can you move all try/except blocks within example body into the __main__ block?
@Tomas2D done, all examples now have the same `__main__` with centralized error handling

python/examples/memory/summarizeMemory.py
```diff
@@ -1,44 +1,44 @@
...
-            "ollama:granite3.1-dense:8b",
-            # ChatModelParameters(temperature=0\),
-        )
-
-        # Create summarize memory instance
-        memory = SummarizeMemory(llm)
+    # Initialize the LLM with parameters
+    llm = ChatModel.from_name(
+        "ollama:granite3.1-dense:8b",
+        # ChatModelParameters(temperature=0\),
```
Tomas2D, line 15: remove `\`

Tomas2D (dismissed):

Tomas2D (commented):
Not sure whether `sys.exit(e.explain())` is a good practise.

ajbozarth:
> Not sure whether sys.exit(e.explain()) is a good practise

This was what was generally suggested on stack overflow, I had doubts at first too, but it works really well in practice

Tomas2D (approved):

Issue #443, opened 2025-03-01
Agent was not able to resolve the task in 0 iterations


cappelaere:
**Describe the bug**
Agent was not able to resolve the task in 0 iterations

**To Reproduce**

1. install beeai with python
2. copy multi_agents.py
3. python3 multiagents.py

**Expected behavior**
was expecting to get weather in NewYork

**Screenshots / Code snippets**
If applicable, add screenshots or code snippets to help explain your problem.

**Set-up:**

- Bee version: latest
- Model provider ollama:granite3.1-dense:8b

**Additional context**

cappelaere:
This is from getting started.  Was hoping to get the example to work!

planetf1:
In addition pytest is currently failing on main with:
```
============================================================ short test summary info ============================================================
FAILED tests/examples/test_examples.py::test_example_execution[agents/bee.py] - SystemExit: 1
FAILED tests/examples/test_examples.py::test_example_execution[agents/granite.py] - SystemExit: 1
FAILED tests/examples/test_examples.py::test_example_execution[agents/simple.py] - SystemExit: 1
FAILED tests/examples/test_examples.py::test_example_execution[backend/providers/ollama.py] - beeai_framework.backend.errors.ChatModelError: Chat Model error
FAILED tests/examples/test_examples.py::test_example_execution[basic.py] - beeai_framework.agents.errors.AgentError: Agent was not able to resolve the task in 0 iterations.
FAILED tests/examples/test_examples.py::test_example_execution[memory/llmMemory.py] - beeai_framework.backend.errors.ChatModelError: Chat Model error
FAILED tests/examples/test_examples.py::test_example_execution[tools/decorator.py] - beeai_framework.agents.errors.AgentError: Agent was not able to resolve the task in 0 iterations.
FAILED tests/examples/test_examples.py::test_example_execution[tools/duckduckgo.py] - beeai_framework.agents.errors.AgentError: Agent was not able to resolve the task in 0 iterations.
FAILED tests/examples/test_examples.py::test_example_execution[tools/openmeteo.py] - beeai_framework.agents.errors.AgentError: Agent was not able to resolve the task in 0 iterations.
FAILED tests/workflows/test_multi_agents.py::test_multi_agents_workflow_basic - beeai_framework.agents.errors.AgentError: Agent was not able to resolve the task in 0 iterations.
[...]

Issue #441, opened 2025-03-01
Examples hitting ToolInputValidationError


ajbozarth:
_Originally posted by @ajbozarth in https://github.com/i-am-bee/beeai-framework/issues/438#issuecomment-2691840877_

The following examples fail intermittently with the a `ToolInputValidationError`

- `agents/simple.py`
- `tools/duckduckgo.py`
- `tools/openmeteo.py`
- `workflows/multi_agents.py`

For example, here's an output from running `agents/simple.py` with the error handling from #438:
```
Agent(thought) 🤖 :  The user wants to know the current weather in Las Vegas. I need to use the OpenMeteoTool for this information.

Agent(tool_name) 🤖 :  OpenMeteoTool
Agent(tool_input) 🤖 :  {'location': 'Las Vegas', 'type': 'current'}
AgentError(beeai_framework.agents.errors): Maximal amount of global retries (0) has been reached.
  ToolInputValidationError(beeai_framework.tools.errors): Tool input validation error
```

As far as I found, all examples that actually call a Tool are failing.

mmurad2:
Related user reported error - let me know if this warrants a different ticket:

> User :bust_in_silhouette: : how is the weather today in new jersey
> Agent(thought) :robot_face: :  The user is asking about the current weather in New Jersey. I need to use the OpenMeteoTool for this information.
> Agent(tool_name) :robot_face: :  OpenMeteoTool
> Agent(tool_input) :robot_face: :  {'location': 'New Jersey', 'date': '2025-03-01'}
> Agent(tool_output) :robot_face: :  Tool input validation error
> HINT: If you're convinced that the input was correct but the tool cannot process it then use a different tool or say I don't know.
> Agent(thought) :robot_face: :  The OpenMeteoTool could not process the date format correctly, even though I formatted 2025-03-01 as ISO 8601. I will try using DuckDuckGo to find out how to interpret the date format and then attempt the OpenMeteoTool again with the correct parameters.
> Agent(tool_name) :robot_face: :  DuckDuckGo
> AgentError(beeai_framework.agents.errors): Maximal amount of global retries (2) has been reached.
>   LinePrefixParserError(beeai_framework.parsers.line_prefix): The generated output does not adhere to the schema.
>   Parsed text line corresponds to a node "tool_name" which is not a start node!

Tomas2D:
Fixed

Issue #442, opened 2025-03-01
The e2e multi agent workflow tests are failing


ajbozarth:
_Originally posted by @ajbozarth in https://github.com/i-am-bee/beeai-framework/issues/438#issuecomment-2691840877_

The e2e multi agent workflow tests `test_example_execution[workflows/multi_agents.py]` and `test_multi_agents_workflow_creation` are failing.

Any workflow with more than one agent added consistently fails due to `RuntimeError: Event loop is closed` which causes `beeai_framework.agents.errors.AgentError: Maximal amount of global retries (0) has been reached.`

This is separate from #441 which is also affecting the `workflows/multi_agents.py` intermittently

planetf1:
Also reported in #443

ajbozarth:
Closing as this was addressed in #554

PR #456, opened 2025-03-03
fix(agents): handle native tool calling and retries

Tomas2D:
- fix tool calling for Granite
- add prompt templates for tool's empty output,  user's empty message, tool input error
- fix empty retry parameter

Ref: #441
Ref: #443
Ref: #428


ajbozarth (dismissed):
I tested this with #438 and it addresses all the issues you tagged.

#442 still needs to be looked at as tests are still failing intermittently and a variety of errors

planetf1 (approved):
I still get a failure as in #443 with this PR. However I think it best to merge and then review

PR #461, opened 2025-03-03
fix: emitter required for tools

michael-desmond:
Adds a required emitter property on tools. All tool implementations now need to implement this property- providing an emitter instance.

I suggest we also change our tool implementations so that name, description and input_schema are implemented in the same manner for consistency. 

I also fixed a bunch of mypy errors in the code that I touched.

- [ ] Tests are included 

### DIFF
--- a/python/beeai_framework/tools/mcp_tools.py
+++ b/python/beeai_framework/tools/mcp_tools.py
@@ -63,11 +63,6 @@ class MCPTool(Tool[MCPToolOutput]):
         self._name = tool.name
         self._description = tool.description or "No available description, use the tool based on its name and schema."
 
-        self._emitter = Emitter.root().child(
-            namespace=["tool", "mcp", self._name],
-            creator=self,
-        )
-
     @property
     def name(self) -> str:
         return self._name
@@ -79,9 +74,11 @@ class MCPTool(Tool[MCPToolOutput]):
     def input_schema(self) -> str:
         return self._tool.inputSchema
 
-    @property
-    def emitter(self) -> Emitter:
-        return self._emitter
+    def create_emitter(self) -> Emitter:
+        return Emitter.root().child(
+            namespace=["tool", "mcp", self.name],
+            creator=self,
+        )
 
     async def _run(self, input_data: Any, options: dict | None = None) -> MCPToolOutput:
         """Execute the tool with given input."""
--- a/python/beeai_framework/tools/mcp_tools.py
+++ b/python/beeai_framework/tools/mcp_tools.py
@@ -74,7 +74,7 @@ class MCPTool(Tool[MCPToolOutput]):
     def input_schema(self) -> str:
         return self._tool.inputSchema
 
-    def create_emitter(self) -> Emitter:
+    def _create_emitter(self) -> Emitter:
         return Emitter.root().child(
             namespace=["tool", "mcp", self.name],
             creator=self,
--- a/python/beeai_framework/tools/mcp_tools.py
+++ b/python/beeai_framework/tools/mcp_tools.py
@@ -63,7 +63,7 @@ class MCPTool(Tool[MCPToolOutput]):
         self._name = tool.name
         self._description = tool.description or "No available description, use the tool based on its name and schema."
 
-        self.emitter = Emitter.root().child(
+        self._emitter = Emitter.root().child(
             namespace=["tool", "mcp", self._name],
             creator=self,
         )
@@ -78,6 +78,10 @@ class MCPTool(Tool[MCPToolOutput]):
 
     def input_schema(self) -> str:
         return self._tool.inputSchema
+
+    @property
+    def emitter(self) -> Emitter:
+        return self._emitter
 
     async def _run(self, input_data: Any, options: dict | None = None) -> MCPToolOutput:
         """Execute the tool with given input."""
--- a/python/beeai_framework/tools/search/base.py
+++ b/python/beeai_framework/tools/search/base.py
@@ -37,4 +37,4 @@ class SearchToolOutput(ToolOutput):
         return len(self.results) == 0
 
     def sources(self) -> list[str]:
-        return {result.url for result in self.results}
+        return [result.url for result in self.results]
--- a/python/beeai_framework/tools/tool.py
+++ b/python/beeai_framework/tools/tool.py
@@ -16,6 +16,7 @@
 import inspect
 from abc import ABC, abstractmethod
 from collections.abc import Callable
+from functools import cached_property
 from typing import Any, Generic, TypeVar
 
 from pydantic import BaseModel, ConfigDict, ValidationError, create_model
@@ -71,7 +72,6 @@ class StringToolOutput(ToolOutput):
 class Tool(Generic[T], ABC):
     def __init__(self, options: dict[str, Any] | None = None) -> None:
         self.options: dict[str, Any] | None = options or None
-        self._emitter: Emitter | None = None
 
     @property
     @abstractmethod
@@ -88,14 +88,12 @@ class Tool(Generic[T], ABC):
     def input_schema(self) -> type[T]:
         pass
 
-    @property
+    @cached_property
     def emitter(self) -> Emitter:
-        if self._emitter is None:
-            self._emitter = self.create_emitter()
-        return self._emitter
+        return self._create_emitter()
 
     @abstractmethod
-    def create_emitter(self) -> Emitter:
+    def _create_emitter(self) -> Emitter:
         pass
 
     @abstractmethod
@@ -213,7 +211,7 @@ def tool(tool_function: Callable) -> Tool:
         def __init__(self, options: dict[str, Any] | None = None) -> None:
             super().__init__(options)
 
-        def create_emitter(self) -> Emitter:
+        def _create_emitter(self) -> Emitter:
             return Emitter.root().child(
                 namespace=["tool", "custom", to_safe_word(self.name)],
                 creator=self,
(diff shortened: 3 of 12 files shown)


python/beeai_framework/tools/weather/openmeteo.py
```diff
@@ -90,8 +94,8 @@ def get_params(self, input: OpenMeteoToolInput) -> dict[str, Any]:
         }
 
         geocode = self._geocode(input)
-        params["latitude"] = geocode.get("latitude")
-        params["longitude"] = geocode.get("longitude")
+        params["latitude"] = geocode.get("latitude") or ""
+        params["longitude"] = geocode.get("longitude") or ""
```
vabarbosa, line 98: couldnt we do?
```suggestion
        params["latitude"] = geocode.get("latitude", "")
        params["longitude"] = geocode.get("longitude", "")
```

python/examples/tools/custom/openlibrary.py
```diff
@@ -59,9 +63,9 @@ async def _run(self, tool_input: OpenLibraryToolInput, _: Any | None = None) ->
             json_output = response.json()[f"{key}:{value}"]
 
         return OpenLibraryToolResult(
-            preview_url=json_output.get("preview_url"),
-            info_url=json_output.get("info_url"),
-            bib_key=json_output.get("bib_key"),
+            preview_url=json_output.get("preview_url") or "",
+            info_url=json_output.get("info_url") or "",
+            bib_key=json_output.get("bib_key") or "",
```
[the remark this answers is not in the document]
michael-desmond: Yes it could, and should :) thanks

python/beeai_framework/tools/tool.py
```diff
@@ -58,14 +69,8 @@ def get_text_content(self) -> str:
...
 class Tool(Generic[T], ABC):
-    options: dict[str, Any]
-
-    emitter: Emitter
-
     def __init__(self, options: dict[str, Any] | None = None) -> None:
-        if options is None:
-            options = {}
-        self.options = options
+        self._options: dict[str, Any] | None = options or None
```
Tomas2D, line 73: Why did you make it a private member?

python/beeai_framework/tools/tool.py
```diff
@@ -189,12 +203,16 @@ class FunctionTool(Tool):
...
-            self.emitter = Emitter.root().child(
+            self._emitter = Emitter.root().child(
                 namespace=["tool", "custom", to_safe_word(self.name)],
                 creator=self,
             )
 
-        async def _run(self, tool_in: Any, _: dict[str, Any] | None = None) -> None:
+        @property
+        def emitter(self) -> Emitter:
+            return self._emitter
```
Tomas2D, line 213: I think that you could just do

```
self.emitter=...
```

and then you don't need to write an external getter all the time.

python/beeai_framework/tools/tool.py
```diff
@@ -189,12 +203,16 @@ class FunctionTool(Tool):
...
+
+        async def _run(self, tool_in: Any, _: ToolRunOptions | None = None) -> None:
             tool_input_dict = tool_in.model_dump()
```
Tomas2D, line 216: `tool_in: Any` should be `input: BaseModel` or ideally a generic

python/beeai_framework/tools/weather/openmeteo.py
```diff
@@ -52,12 +52,16 @@ class OpenMeteoTool(Tool[OpenMeteoToolInput]):
...
-        self.emitter = Emitter.root().child(
+        self._emitter = Emitter.root().child(
             namespace=["tool", "weather", "openmeteo"],
             creator=self,
         )
 
-    def _geocode(self, input: OpenMeteoToolInput) -> dict[str, Any]:
+    @property
+    def emitter(self) -> Emitter:
+        return self._emitter
```
Tomas2D, line 62: same

Tomas2D (changes requested):

python/beeai_framework/tools/tool.py
michael-desmond: Prior to this it was a class instance variable, shadowed by an instance variable, which could lead to it being overwritten when using multiple tools. Afaik this should be a instance variable? We can add a getter if it needs to be public.

python/beeai_framework/tools/tool.py
michael-desmond: The external getter is just a way to force the tool developer to provide an emitter. Its a abstract property on the base Tool class. If you try to create a tool without implememeting the getter it will fail.

python/beeai_framework/tools/tool.py
michael-desmond: Ill fix this.

python/beeai_framework/tools/tool.py
michael-desmond: Added abstract `create_emitter` method to tool base.

ajbozarth (approved):
great work this LGTM

I believe we need to do the same for BaseAgent, ChatModel, and Workflow, can you open follow up issues for addressing those with this same fix?

python/beeai_framework/tools/mcp_tools.py
```diff
@@ -79,6 +74,12 @@ def description(self) -> str:
     def input_schema(self) -> str:
         return self._tool.inputSchema
 
+    def create_emitter(self) -> Emitter:
+        return Emitter.root().child(
+            namespace=["tool", "mcp", self.name],
+            creator=self,
+        )
```
Tomas2D, line 81: Will this work?

`create_emitter` gets called in the parent constructor, but you are setting `self._name=tool.name` on line 63

Tomas2D: But because the tool itself is not working, we can ignore it.

Tomas2D: Or maybe you could easily fix that by removing `self._name` and `self._description` and adding these:

```python
@property
def name():
   return self._tool.name

@property
def description():
   return self._tool.description or "No available description, use the tool based on its name and schema."
```

python/beeai_framework/tools/search/base.py
```diff
@@ -37,4 +37,4 @@ def is_empty(self) -> bool:
         return len(self.results) == 0
 
     def sources(self) -> list[str]:
-        return {result.url for result in self.results}
+        return [result.url for result in self.results]
```
Tomas2D, line 40: 👍🏻

python/beeai_framework/tools/tool.py
```diff
@@ -82,8 +88,18 @@ def description(self) -> str:
     def input_schema(self) -> type[T]:
         pass
 
+    @property
+    def emitter(self) -> Emitter:
+        if self._emitter is None:
+            self._emitter = self.create_emitter()
+        return self._emitter
```
Tomas2D, line 95: What about using `cached` decorator?

python/beeai_framework/tools/tool.py
```diff
@@ -58,14 +69,9 @@ def get_text_content(self) -> str:
...
-    options: dict[str, Any]
-
-    emitter: Emitter
-
     def __init__(self, options: dict[str, Any] | None = None) -> None:
-        if options is None:
-            options = {}
-        self.options = options
+        self.options: dict[str, Any] | None = options or None
+        self._emitter: Emitter | None = None
```
Tomas2D, line 74: what about `self.emitter = self.create_emitter()`

python/beeai_framework/tools/tool.py
```diff
@@ -82,8 +88,18 @@ def description(self) -> str:
...
         pass
 
+    @property
+    def emitter(self) -> Emitter:
+        if self._emitter is None:
+            self._emitter = self.create_emitter()
+        return self._emitter
+
+    @abstractmethod
+    def create_emitter(self) -> Emitter:
```
Tomas2D, line 98: Should be `_create_emitter` as we don't want to prevent our users from the unintended invocation

Tomas2D (changes requested):

python/beeai_framework/tools/mcp_tools.py
michael-desmond: I fixed this is a more general way by lazily creating the emitter when the emitter property is called on the superclass. This means we dont need to worry about order of instantiation in the subclass, or at least the potential problem is more obvious.

python/beeai_framework/tools/tool.py
michael-desmond: If we do this then there is the possibility that a subclass will end up using undefined variables when the emitter is created. The MCPTool is an example of this situation. Using lazy loading via property gets around this issue without requiring the subclass to know about it.

python/beeai_framework/tools/tool.py
michael-desmond: ok makes sense.

python/beeai_framework/tools/tool.py
michael-desmond: Yes @cached_property would further simplify the code, good suggestion

python/beeai_framework/tools/tool.py
Tomas2D: Got it, but I would say that we would like to fail ASAP and not during a runtime.

Tomas2D: If the name becomes invalid, the emitter will complain.

michael-desmond: After implementing the @cached_property we no longer need this variable definition.

Tomas2D (approved):

PR #467, opened 2025-03-04
chore: update release script

Tomas2D:
Verify that the local branch is up to sync with the origin before releasing.


vabarbosa (approved):
👍🏽

Issue #453, opened 2025-03-03
Chat model parameters are not being used


Tomas2D:
A user can set default parameters for the chat model instance via the `config` method; however, these parameters are not used anywhere.

- [x] start using `parameters` instead of `settings` in LiteLLM
- [x] set default temperature to 0
- [ ] correctly merge all parameters before calling litellm completion method

PR #486, opened 2025-03-05
fix(backend): correctly merge inference parameters

Tomas2D:
Ref: #453


planetf1 (commented):
This breaks groq - issue raised #487 (due to additional paramers being ignored)

planetf1 (approved):

PR #494, opened 2025-03-05
feat(examples): adds logger examples and updates docs

AngeloDanducci:

Ref: #336 

### Description

Adds logger examples and updates docs

### Checklist

- [x] Tests are included 

### DIFF
--- a/python/docs/logger.md
+++ b/python/docs/logger.md
@@ -45,8 +45,23 @@ In the BeeAI framework, the `Logger` class is an abstraction built on top of Pyt
 
 To use the logger in your application:
 
+<!-- embedme examples/logger/base.py -->
 ```py
-# Coming soon
+import logging
+
+from beeai_framework.utils import BeeLogger
+
+# Configure logger with default log level
+logger = BeeLogger("app", level=logging.TRACE)
+
+# Log at different levels
+logger.trace("Trace!")
+logger.debug("Debug!")
+logger.info("Info!")
+logger.warning("Warning!")
+logger.error("Error!")
+logger.fatal("Fatal!")
+
 ```
 
 _Source: examples/logger/base.py_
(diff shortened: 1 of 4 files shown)


python/docs/logger.md
```diff
@@ -45,8 +45,23 @@ In the BeeAI framework, the `Logger` class is an abstraction built on top of Pyt
...
-# Coming soon
+import logging
+
+from beeai_framework.utils import BeeLogger
+
+# Configure logger with default log level
+logger = BeeLogger("app", level=logging.TRACE)
+
+# Log at different levels
+logger.trace("Trace!")
```
Tomas2D, line 58: Can we add typings for a custom level? My IDE complains that there is no such method.

python/examples/logger/agent.py
```diff
@@ -0,0 +1,37 @@
+import asyncio
```
Tomas2D, line 1: This example does not bring any value. I would remove it.

Tomas2D (changes requested):

python/docs/logger.md
ajbozarth: I did some looking into this and Mypy just doesn't support `setattr` since it is static analysis and `setattr` is dynamically set at runtime. So in this case we would want to just exempt this from Mypy, but adding a Mypy ignore in an example that's embedded in the docs would look bad.

python/examples/logger/agent.py
ajbozarth: This example exists in TS and is embedded in the docs for both TS and python. So removing it would lower the quality of the docs. (I do agree it's very redundant with other agent examples though)

python/docs/logger.md
Tomas2D: we can just add such method to the BeeLogger

ajbozarth: As far as I found online that's non-trivial and I believe the dynamic solution we're using is actually the recommend way of adding logging levels. If want to revisit it I would suggest we open up a follow up issue and merge this as is since it works as intend despite the Mypy complaints

ajbozarth (approved):

python/docs/logger.md
Tomas2D: I meant this.

```python
from typing import TYPE_CHECKING

class BeeLogger(logging.Logger):
    if TYPE_CHECKING:
        trace = logging.Logger.debug
```

Tomas2D: This satisfies MyPy without any modifications.

Tomas2D (approved):
Looks good 👍🏻

PR #498, opened 2025-03-06
test(examples): Add missing backend examples

ajbozarth:

Ref #336 

### Description

Adds the basic backend examples and updates the docs. Also includes some Mypy fixes found during my work.

### Checklist

- [x] Tests are included 

### DIFF
--- a/python/beeai_framework/tools/tool.py
+++ b/python/beeai_framework/tools/tool.py
@@ -31,7 +31,6 @@ from beeai_framework.utils.strings import to_safe_word
 
 logger = BeeLogger(__name__)
 
-R = TypeVar("R", bound=BaseModel)
 IN = TypeVar("IN", bound=BaseModel)
 
 
@@ -101,7 +100,7 @@ class Tool(Generic[IN, OPT], ABC):
         pass
 
     @abstractmethod
-    async def _run(self, input: IN, options: OPT | None, context: RunContext) -> R:
+    async def _run(self, input: IN, options: OPT | None, context: RunContext) -> Any:
         pass
 
     def validate_input(self, input: IN | dict[str, Any]) -> IN:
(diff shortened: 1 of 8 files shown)


python/beeai_framework/tools/tool.py
```diff
@@ -31,6 +31,7 @@
 
 logger = BeeLogger(__name__)
 
+R = TypeVar("R", bound=BaseModel)
```
michael-desmond, line 34: This should be OUT, we already have IN and OPT. R is a little ambiguous. I think it should also be bound to ToolOutput rather than BaseModel.

python/beeai_framework/tools/tool.py
```diff
@@ -100,7 +101,7 @@ def _create_emitter(self) -> Emitter:
         pass
 
     @abstractmethod
-    async def _run(self, input: IN, options: OPT | None, context: RunContext) -> Any:
+    async def _run(self, input: IN, options: OPT | None, context: RunContext) -> R:
```
michael-desmond, line 104: All of our Tool implementation (openmeteo etc) should now be instantiated with the appropriate concrete types. ie. `class OpenMeteoTool(Tool[OpenMeteoToolInput, ToolRunOptions, StringToolOutput]):`

michael-desmond (changes requested):

python/beeai_framework/tools/tool.py
ajbozarth: I tried to do this and hit an issue with Run expecting a BaseModel output, so I think this is better handled in a separate PR, I only added these to fix some mypy issues. I'm going to revert the non-example changes to unblock this

ajbozarth: pushed the revert, this PR will just leave the Tool class as is and we can sort out those fixes in a PR not focused on adding backend examples

Tomas2D (approved):
I would not set temperature=0

PR #481, opened 2025-03-05
feat(tools): add MCP tool

akihikokuroda:
### Which issue(s) does this pull-request address?

#398 

This PR adds MCPTool and 2 examples.

### Checklist

- [x] I have read the [contributor guide](https://github.com/i-am-bee/beeai-framework/blob/main/CONTRIBUTING.md)
- [x] I have [signed off](https://github.com/i-am-bee/beeai-framework/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-dco) on my commit
- [x] Linting passes: `yarn lint` or `yarn lint:fix`
- [x] Formatting is applied: `yarn format` or `yarn format:fix`
- [x] Unit tests pass: `yarn test:unit`
- [ ] E2E tests pass: `yarn test:e2e`
- [ ] Tests are included 
- [ ] Documentation is changed or added
- [x] Commit messages and PR title follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary)

PR #499, opened 2025-03-06
fix: improve decorated tool output

michael-desmond:
1. Replaces to_string on tool output with python __str__ override
2. Tool results are now printed in agents notebook
3. Decorated tools that return str can now work with Agent (output is wrapped as StringToolOutput)
4. Error checking in notebooks for search tool failures

### Checklist

- [ ] Tests are included 

### DIFF
--- a/python/beeai_framework/tools/tool.py
+++ b/python/beeai_framework/tools/tool.py
@@ -60,6 +60,9 @@ class ToolOutput(ABC):
         return self.get_text_content()
 
 
+OUT = TypeVar("OUT", bound=ToolOutput)
+
+
 class StringToolOutput(ToolOutput):
     def __init__(self, result: str = "") -> None:
         super().__init__()
@@ -83,7 +86,7 @@ class JSONToolOutput(ToolOutput):
         return not self.result
 
 
-class Tool(Generic[IN, OPT], ABC):
+class Tool(Generic[IN, OPT, OUT], ABC):
     def __init__(self, options: dict[str, Any] | None = None) -> None:
         self.options: dict[str, Any] | None = options or None
 
@@ -111,7 +114,7 @@ class Tool(Generic[IN, OPT], ABC):
         pass
 
     @abstractmethod
-    async def _run(self, input: IN, options: OPT | None, context: RunContext) -> Any:
+    async def _run(self, input: IN, options: OPT | None, context: RunContext) -> OUT:
         pass
 
     def validate_input(self, input: IN | dict[str, Any]) -> IN:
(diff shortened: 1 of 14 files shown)


python/beeai_framework/tools/tool.py
```diff
@@ -220,12 +220,12 @@ def _create_emitter(self) -> Emitter:
                 creator=self,
             )
 
-        async def _run(self, input: Any, options: ToolRunOptions | None, context: RunContext) -> Any:
+        async def _run(self, input: Any, options: ToolRunOptions | None, context: RunContext) -> StringToolOutput:
```
vabarbosa, line 223: i think the return type hint for the tool _run should probably be ToolOutput especially with the JSONToolOutput is incoming: https://github.com/i-am-bee/beeai-framework/pull/481/files#diff-bbeaddb800a3c89929ddeda85a1c3bfc830376459e30c308c667412197d5ec04R73-R83

ajbozarth: I was about to say this, I think the tool decorator shouldn't be restricted to just string output

Tomas2D: We should either use generic `ToolOutput` or ideally introduce a generic for that.

python/beeai_framework/tools/tool.py
```diff
@@ -220,12 +220,12 @@ def _create_emitter(self) -> Emitter:
...
             tool_input_dict = input.model_dump()
             if inspect.iscoroutinefunction(tool_function):
-                return await tool_function(**tool_input_dict)
+                return StringToolOutput(str(await tool_function(**tool_input_dict)))
             else:
-                return tool_function(**tool_input_dict)
+                return StringToolOutput(str(tool_function(**tool_input_dict)))
```
Tomas2D, line 228: this should not be needed

michael-desmond: We added this because if you try to create a decorated tool without returning a ToolOutput then the tool wont work with an agent. Alternative is to change agent code when handling tool outputs.

Tomas2D: In that case, we should conditionally check if the return is an instance of ToolOutput and if not, we should wrap it.

python/beeai_framework/tools/tool.py
```diff
@@ -231,12 +234,17 @@ def _create_emitter(self) -> Emitter:
...
-                return await tool_function(**tool_input_dict)
+                result = await tool_function(**tool_input_dict)
+            else:
+                result = tool_function(**tool_input_dict)
+
+            if isinstance(result, ToolOutput):
+                return result
             else:
-                return tool_function(**tool_input_dict)
+                return StringToolOutput(result=str(result))
```
vabarbosa, line 247: should this also check if result is dict then use JSONToolOutput else default to StringToolOutput?

python/beeai_framework/tools/weather/openmeteo.py
```diff
@@ -46,7 +46,7 @@ class OpenMeteoToolInput(BaseModel):
     )
 
 
-class OpenMeteoTool(Tool[OpenMeteoToolInput, ToolRunOptions]):
+class OpenMeteoTool(Tool[OpenMeteoToolInput, ToolRunOptions, StringToolOutput]):
```
vabarbosa, line 49: now that we have JSONToolOutput isnt the output from OpenMeteo more accurately a JSONToolOutput instead of StringToolOutput?

Issue #398, opened 2025-02-25
MCP integration with Python Library


mmurad2:
**Description**
Ability to use MCP tools with Python framework

**Acceptance criteria**
Parity with TS implementation, docs and examples: https://github.com/i-am-bee/beeai-framework/blob/main/typescript/docs/tools.md#using-the-mcptool-class

Blocked by: #337

Tomas2D:
@akihikokuroda did the initial implementation that is blocked by #337, let's finish it first and then @akihikokuroda can finish that.

akihikokuroda:
The `_run` method of the Tool is not `async` yet.  It is required for the MCP tool implementation and the change affects all tools.

vabarbosa:
> The `_run` method of the Tool is not `async` yet. It is required for the MCP tool implementation and the change affects all tools.

incoming https://github.com/i-am-bee/beeai-framework/pull/452

PR #502, opened 2025-03-06
feat(internals): construct Pydantic model from JSON Schema

Tomas2D:
Creates more robust Pydantic model based on provided JSON Schema

Ref: #398


akihikokuroda (approved):
LGTM

PR #504, opened 2025-03-06
refactor: move logger to the root

Tomas2D:
Match TS directory structure.


vabarbosa:
i believe the tools.md and logger.md may require updating accordingly

Tomas2D:
Good catch. Updated.

vabarbosa (approved):

Issue #434, opened 2025-02-28
add mypy to CI


jenna-winkler:

vabarbosa:
thanks @adigidh!

if possible and it makes sense, maybe you can add it as a `poe` task e.g.,: https://github.com/i-am-bee/beeai-framework/blob/main/python/pyproject.toml#L136-L145
and then include/run the task as part of the existing GH action e.g.,: https://github.com/i-am-bee/beeai-framework/blob/main/.github/workflows/py_main.yml#L35-L38

and yes, for now we would want to allow the workflow to continue even if there are `mypy` errors. if we get to a point where we have addressed all the failures then we can revisit the idea of having any new `mypy` errors failing the workflow.

thank you for you interest in assisting with this!

PR #507, opened 2025-03-06
434: Static Type Checker Mypy added to CI & as a poe task in pyproject.toml

adigidh:

Closes: https://github.com/i-am-bee/beeai-framework/issues/434

### Description

Added MyPy type checking as a poe task and as a CI workflow action. Added in tasks for different type checking scenarios.

type-check = "mypy ."
type-check-strict = "mypy --strict ."
type-check-verbose = "mypy --verbose --show-error-context --pretty ."

Captured examples of failures in the Git Actions: https://github.com/i-am-bee/beeai-framework/actions/runs/13707535290/job/38336396042?pr=507

Example failures while running these locally: `poetry run poe type-check-strict`
image

### Checklist

- [ ] Linting passes: `yarn lint` or `yarn lint:fix`
- [ ] Formatting is applied: `yarn format` or `yarn format:fix`
- [ ] Unit tests pass: `yarn test:unit`
- [ ] Tests are included 


ajbozarth:
This looks great, would you be willing to look into addressing the top level mypy failures while you're at it. We should be able to config mypy in pyproject.toml to allow duplicate module names in separate subdirectories.

adigidh:
> This looks great, would you be willing to look into addressing the top level mypy failures while you're at it. We should be able to config mypy in pyproject.toml to allow duplicate module names in separate subdirectories.

@ajbozarth : Initially I updated mypy’s configuration to allow for duplicate module names in separate subdirectories.

```
namespace_packages = true: Allows mypy to recognize namespaces without requiring __init__.py files in every directory.

explicit_package_bases = true: Ensures mypy uses the directory structure (relative to the root).
```

But it complained about other things with “Source file found twice under different module names". Therefore added empty `__init__.py` files to enforce a namespace, offerings a cleanest way to fix both issues spotted earlier. Let me know what you think!

ajbozarth:
> But it complained about other things with "Source file found twice under different module names". Therefore added empty __init__.py files to enforce a namespace, offering a cleanest way to fix both issues spotted earlier. Let me know what you think!

Thanks, I checked it out and played with it a bit and I think you only need the __init__.py at the root of examples to fix it, so you can delete the others. Otherwise I think this is g2g

python/pyproject.toml
```diff
@@ -245,6 +245,12 @@ help = "Extract Python files only"
 args = [{ name = "TARGET", positional = true, multiple = true }]
 shell = "printf '%s\\n' $TARGET | grep -E '\\.(py|ipynb)$' | tr '\\n' ' '"
 
+[tool.poe.tasks]
```
ajbozarth, line 248: I think this should work like poe Lin and poe format, so this would be `[tool.poe.tasks.type-check]` and reformat the following lines to work similar to the other poe.tasks section above

adigidh: sure thing, updated the configs and the added docs.

ajbozarth (approved):
This looks fantastic, thank you for the contribution

Issue #512, opened 2025-03-06
Granite sample - better handling of inappropriate questions


planetf1:
In the following example we ask for the weather without specifying a location.
The tool fails without a location, but the handling of this within the agent needs improving

```
(beeai-framework-py3.13) ➜  python git:(azureopenai) ✗ python examples/agents/granite.py
Interactive session has started. To escape, input 'q' and submit.
User 👤 : what is the weather today
Agent(thought) 🤖 :  The user wants to know the current weather. I will use the OpenMeteoTool with the location_name parameter set to the user's current location and temperature_unit set to celsius.

Agent(tool_name) 🤖 :  OpenMeteoTool
Agent(tool_input) 🤖 :  {'location_name': 'current location', 'temperature_unit': 'celsius'}
Agent(tool_output) 🤖 :  The tool has failed; the error log is shown below. If the tool cannot accomplish what you want, use a different tool or explain why you can't use it.

ToolError(beeai_framework.tools.errors): Tool Error
  KeyError(builtins): 'results'
Agent(thought) 🤖 :  The OpenMeteoTool failed to retrieve the weather information. I will ask the user for their current location and try again.

Agent(final_answer) 🤖 :  I'm sorry, but I couldn't get the weather information for your current location. Could you please provide me with a city or town name so I can fetch the weather data?
Agent 🤖 :  I'm sorry, but I couldn't get the weather information for your current location. Could you please provide me with a city or town name so I can fetch the weather data?
```

Tomas2D:
What would you expect?

planetf1:
Originated in [slack thread](https://ibm-research.slack.com/archives/C08DGUBRU3B/p1741278948904319)

The thought here was about the handling in the agent - may be a combination of prompt and any underlying tool issues (should be addressed now). In the above example the agent did not prompt for location, which was expected.

The reason is that this example doesn't loop back for more user input, whilst other examples do.

Could consider whether we want consistency here?

Tomas2D:
Could you verify @planetf1 ?

planetf1:
The issue is fixed here, but the ask is to update the granite sample to loop on user input, similarly to the regular examples.

As part of this it's worth checking against the other examples

PR #513, opened 2025-03-06
fix(tools): handle non-existing locations in OpenMeteo

Tomas2D:
Ref: #512

### DIFF
--- a/python/beeai_framework/tools/weather/openmeteo.py
+++ b/python/beeai_framework/tools/weather/openmeteo.py
@@ -26,7 +26,7 @@ from pydantic import BaseModel, Field
 from beeai_framework.context import RunContext
 from beeai_framework.emitter.emitter import Emitter
 from beeai_framework.logger import Logger
-from beeai_framework.tools import ToolInputValidationError
+from beeai_framework.tools import ToolError, ToolInputValidationError
 from beeai_framework.tools.tool import StringToolOutput, Tool, ToolRunOptions
 
 logger = Logger(__name__)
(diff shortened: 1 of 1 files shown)


python/beeai_framework/tools/weather/openmeteo.py
```diff
@@ -75,7 +75,9 @@ def _geocode(self, input: OpenMeteoToolInput) -> dict[str, str]:
         )
 
         response.raise_for_status()
-        results = response.json()["results"]
+        results = response.json().get("results", [])
+        if not results:
+            raise ToolError(f"Location '{input.location_name}' was not found.")
```
ajbozarth, line 80: Is this really the only possible error case if results is empty?

ajbozarth (approved):
Code looks good, just one clarifying question

python/beeai_framework/tools/weather/openmeteo.py
Tomas2D: `keyerror` error is not very descriptive, this prevents that

planetf1 (approved):

python/beeai_framework/tools/weather/openmeteo.py
Tomas2D: yes

Tomas2D: If you call the API with the location that does not exist, then you get an object without the `results` attribute. 
I added a check that fixes that.

Issue #471, opened 2025-03-04
We should add a new `Run.on` API endpoint


ajbozarth:
### Discussed in https://github.com/orgs/i-am-bee/discussions/386

<div type='discussions-op-text'>

<sup>Originally posted by **ajbozarth** February 24, 2025</sup>
This idea came out of some user feedback and the same discussion that brought up #385
I'll be using python in my examples, but this would be an update to both python and typescript.

Currently if a user wants to add a callback on a certain event they are required to nest two callbacks: An a observer callback that takes in an emitter and listens on that emitter for their desired event and the event callback that is then called with that event and it's data. This use case is fine when the observer function might want to do multiple thing and not just one `emitter.on`, but in the more common case with only one listener this is overly verbose, as seen in the following example:

```python
response = await agent.run(
    BeeRunInput(prompt="What's the current weather in Las Vegas?")
).observe(lambda emitter: emitter.on("success", lambda data, event: print(data.result)))
```

I propose we add a new function `Run.on` (or `Run.observeOn`) that combines the `Run.observe` and `Emitter.on` calls. In conjunction with #385 this would simplify the above example to the following:

```python
response = await agent.run("What's the current weather in Las Vegas?").on("success", lambda data, event: print(data.result))
```

On suggestion of @Tomas2D I have created this as a discussion since this could be considered a significant API change. </div>

PR #516, opened 2025-03-06
feat: Add Run.on API to directly add an event callback

ajbozarth:

Closes: #471

### Description

This introduces the `Run.on` method that combines the `Run.observe` and `Emitter.match` calls into one more user friendly callback.

Included in this PR:
- The `Run.on` method and rework of `Run` internals to support it
- Updates to a few demos to leverage the new method
- Address mypy errors by moving the `ContextVar` storage to a global, as a `ContextVar` is intended to be

### Checklist

- [x] Tests are included 

### DIFF
--- a/python/beeai_framework/context.py
+++ b/python/beeai_framework/context.py
@@ -25,7 +25,7 @@ from typing import Any, Generic, Self, TypeVar
 from pydantic import BaseModel
 
 from beeai_framework.cancellation import AbortController, AbortSignal, register_signals
-from beeai_framework.emitter import Emitter, EventTrace
+from beeai_framework.emitter import Callback, Emitter, EmitterOptions, EventTrace, Matcher
 from beeai_framework.errors import AbortError, FrameworkError
 from beeai_framework.logger import Logger
 from beeai_framework.utils.asynchronous import ensure_async
@@ -33,6 +33,8 @@ from beeai_framework.utils.asynchronous import ensure_async
 R = TypeVar("R")
 
 logger = Logger(__name__)
+
+storage: ContextVar["RunContext"] = ContextVar("storage")
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


ajbozarth:
There can still be discussion around the new method's name. I went with `Run.on` but I also considered the following options if there's strong opinion to switch to one of those: `observe_on`, `match`, `observe_match`, `on_event`, or to try and overload `observe` itself (not trivial)

python/beeai_framework/context.py
```diff
@@ -80,8 +86,6 @@ def _set_context(self, context: dict) -> None:
 
 
 class RunContext(RunInstance):
-    storage: ContextVar[Self] = ContextVar("storage", default=None)
```
Tomas2D, line 83: Why did you remove it? It seems more straightforward to have it there as you are using it in the class.

python/beeai_framework/context.py
```diff
@@ -49,27 +51,31 @@ class Run(Generic[R]):
...
+        self.tasks.append((fn, [self.run_context]))
         return self
 
     async def _run_tasks(self) -> R:
-        for fn, param in self.tasks:
-            await ensure_async(fn)(param)
+        for fn, params in self.tasks:
+            await ensure_async(fn)(*params)
 
         self.tasks.clear()
```
Tomas2D, line 80: clear should be done in advance

python/beeai_framework/context.py
ajbozarth: mypy was upset we were referencing an instance var from a static method. Then when I looked into some fixes for that I found that ContextVar are designed be global vars not instance vars

python/beeai_framework/context.py
ajbozarth: What do you mean? clear() empties the list and we iterate over the list in the chunk above this. Where are we supposed the do it?

Tomas2D:
```
tasks = self.tasks.splice()
for fn, params in tasks:
  await ensure_async(fn)(*params)
```

Tomas2D: but it should be ok as it is now

ajbozarth: I think I understand, since the callbacks can be asynchronous you want to copy the task list, clear self.list, then run the tasks from the copy:
```python
    async def _run_tasks(self) -> R:
        tasks = self.tasks[:]
        self.tasks.clear()

        for fn, params in tasks:
            await ensure_async(fn)(*params)

        return await self.handler()
```

ajbozarth: done

Tomas2D (approved):

PR #519, opened 2025-03-07
feat: improve error messages

Tomas2D:


typescript/src/errors.ts
```diff
@@ -93,9 +93,7 @@ export class FrameworkError extends AggregateError {
   }
 
   static ensure(error: Error): FrameworkError {
-    return error instanceof FrameworkError
-      ? error
-      : new FrameworkError("Framework error has occurred.", [error]);
+    return error instanceof FrameworkError ? error : new FrameworkError(undefined, [error]);
```
GALLLASMILAN, line 96: Why do you return undefined instead of the specific error message here?

GALLLASMILAN (approved):

PR #523, opened 2025-03-07
fix(backend): handle event loop closed error

Tomas2D:
Ref: #442


vabarbosa (approved):
nice 👍🏽

ajbozarth (approved):
tested locally, LGTM

Issue #532, opened 2025-03-07
Add support for extra headers to be sent using openai backend


planetf1:
Some backend providers supporting the openai API require additional
HTTP headers, for example to provide a token.

Add this to our OpenAI provider

PR #533, opened 2025-03-07
feat(adapters): Support additional headers for OpenAI backend

planetf1:

Closes: #532

### Description

[ Updated with revised format / variable ]

Adds support for LiteLLM provider to add additional headers
```
export OPENAI_EXTRA_HEADERS="MY_API_KEY=ab592c9f3e001b"
```

To test (for example, with OpenAI, also set:
```
OPENAI_API_KEY=my-api-key
OPENAI_CHAT_MODEL=my-model
OPENAI_API_BASE=https://my-endpoint/
```

Tested ok

### Checklist

- [ ] Tests are included 

### DIFF
--- a/python/beeai_framework/adapters/litellm/chat.py
+++ b/python/beeai_framework/adapters/litellm/chat.py
@@ -78,6 +78,11 @@ class LiteLLMChatModel(ChatModel, ABC):
                 if isinstance(parsed_headers, dict):
                     extra_headers = parsed_headers
             except json.JSONDecodeError:
+                logger.error(
+                    f"Error decoding LITELLM_EXTRA_HEADERS: {extra_headers_json}. "
+                    + "This must be a valid JSON string. No extra headers will be set."
+                )
+                # consider raising an exception.            except json.JSONDecodeError:
                 pass
 
         if extra_headers:
--- a/python/beeai_framework/adapters/litellm/chat.py
+++ b/python/beeai_framework/adapters/litellm/chat.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import json
 import logging
 import os
 from abc import ABC
@@ -70,17 +69,25 @@ class LiteLLMChatModel(ChatModel, ABC):
 
         # Extra headers logic
         extra_headers = None
-        extra_headers_json = self._settings.get("extra_headers", os.getenv("LITELLM_EXTRA_HEADERS"))
+        extra_headers_str = self._settings.get("extra_headers", os.getenv("LITELLM_EXTRA_HEADERS"))
 
-        if extra_headers_json:
+        if extra_headers_str:
             try:
-                parsed_headers = json.loads(extra_headers_json)
-                if isinstance(parsed_headers, dict):
+                parsed_headers = {}
+                for pair in extra_headers_str.split(","):
+                    pair = pair.strip()
+                    if "=" in pair:
+                        key, value = pair.split("=", 1)
+                        parsed_headers[key.strip()] = value.strip()
+
+                if parsed_headers:
                     extra_headers = parsed_headers
-            except json.JSONDecodeError:
+
+            except Exception as e:
                 logger.error(
-                    f"Error decoding LITELLM_EXTRA_HEADERS: {extra_headers_json}. "
-                    + "This must be a valid JSON string. No extra headers will be set."
+                    f"Error parsing LITELLM_EXTRA_HEADERS: {extra_headers_str}. "
+                    f"Error: {e}. "
+                    "This must be a comma-separated list of key=value pairs. No extra headers will be set."
                 )
                 # consider raising an exception.            except json.JSONDecodeError:
                 pass
--- a/python/beeai_framework/adapters/litellm/chat.py
+++ b/python/beeai_framework/adapters/litellm/chat.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import logging
-import os
 from abc import ABC
 from collections.abc import AsyncGenerator
 from typing import Any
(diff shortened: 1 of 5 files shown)


python/beeai_framework/adapters/openai/backend/chat.py
```diff
@@ -30,6 +31,21 @@ def provider_id(self) -> ProviderName:
     def __init__(self, model_id: str | None = None, settings: dict | None = None) -> None:
         _settings = settings.copy() if settings is not None else {}
 
+        # Extra headers -- ignored if invalid
+        extra_headers = None
+        extra_headers_json = _settings.get("extra_headers", os.getenv("OPENAI_EXTRA_HEADERS"))
```
vabarbosa, line 36: `extra_headers` is supported across multiple services, e.g.,
- https://docs.litellm.ai/docs/providers/anthropic#passing-extra-headers-to-anthropic-api
- https://docs.litellm.ai/docs/providers/bedrock#passing-extra-headers--custom-api-endpoints

to that end should this really be `OPENAI_EXTRA_HEADERS` and should it be in openai adapter? i would think we want this in the LiteLLM adapter so it could be applicable across other adapters (and replace the `OPENAI_` prefix with something more generic)

python/beeai_framework/adapters/openai/backend/chat.py
```diff
@@ -30,6 +31,21 @@ def provider_id(self) -> ProviderName:
...
+        if extra_headers_json:
+            try:
+                parsed_headers = json.loads(extra_headers_json)
+                if isinstance(parsed_headers, dict):
+                    extra_headers = parsed_headers
+            except json.JSONDecodeError:
+                pass
+
+        if extra_headers:
+            _settings["headers"] = extra_headers
```
vabarbosa, line 47: LiteLLM also accepts `extra_headers` params: 
image

so do we have to copy it over to `headers`? we end up with both `headers` and `extra_headers` in the `_settings`

vabarbosa (changes requested):
thanks... just a couple comments. your thoughts are appreciated

python/beeai_framework/adapters/openai/backend/chat.py
planetf1: it probably should be extra_headers so that we just augment what is already there?

planetf1: Yes, I'd agree - if we can do this at the litellm level it will be much more flexible

planetf1:
I can look at this tomorrow to refactor to make more generic. (if urgent ok to merge than update after)

python/beeai_framework/adapters/openai/backend/chat.py
planetf1: Yes, I'd agree - if we can do this at the litellm level it will be much more flexible

planetf1:
Rather than adding multiple environment variables and managing precedence across inheritance, are we ok to just add a single one?

Previously we had `OPENAI_EXTRA_HEADERS` (in the above working code)

If this is set at the LiteLLM adapter level it might be `LIGHTLLM_EXTRA_HEADERS`

It might be more generic in future if we use `BEEAI_EXTRA_HEADERS` or `BEEAI_LLM_EXTRA_HEADERS` ... ? But the implementation would then differ, and potentially be incomplete.

My thought is to just have a single variable, to name it `BEEAI_LLM_EXTRA_HEADERS`, but only to implement it in LiteLLM adapter

planetf1:
Moved to draft whilst refactoring/testing. Will move back to ready once all comments are acted on and tests complete.

python/beeai_framework/adapters/openai/backend/chat.py
planetf1: Now moved to `LITELLM_EXTRA_HEADERS` and implemented in the litellm adapter

python/beeai_framework/adapters/openai/backend/chat.py
planetf1: Reviewing litellm docs/search : summary:
```
headers: These are the standard HTTP headers that litellm uses for authentication and content negotiation. For example, litellm automatically sets headers like Content-Type: application/json and Accept: application/json. You can also use the headers parameter to override or add to these standard headers. A common use case is to set an Authorization header with your API key if needed for certain model providers.

extra_headers:  This parameter is designed for passing custom headers that are not part of the standard HTTP headers used by litellm.  extra_headers is particularly useful when you are routing requests through an API management gateway or a reverse proxy and need to include specific headers for authorization, project identification, or other custom logic handled by the gateway. The value of extra_headers should be a JSON string representing the headers you want to add.
```
For this use case the latter seems more appropriate - so updated accordingly

python/beeai_framework/adapters/openai/backend/chat.py
```diff
@@ -30,6 +31,21 @@ def provider_id(self) -> ProviderName:
...
+        extra_headers = None
+        extra_headers_json = _settings.get("extra_headers", os.getenv("OPENAI_EXTRA_HEADERS"))
+
+        if extra_headers_json:
+            try:
+                parsed_headers = json.loads(extra_headers_json)
+                if isinstance(parsed_headers, dict):
+                    extra_headers = parsed_headers
+            except json.JSONDecodeError:
+                pass
```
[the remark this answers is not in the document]
planetf1: Given the other points about moving to litellm (which many other (all?) adapters are based on currently I think I'd defer further refactor until we identify an additional need for this outside litellm

planetf1:
Ready for review

python/beeai_framework/adapters/litellm/chat.py
```diff
@@ -55,14 +58,39 @@ class LiteLLMChatModel(ChatModel, ABC):
...
+
+        # Extra headers logic
+        extra_headers = None
+        extra_headers_json = self._settings.get("extra_headers", os.getenv("LITELLM_EXTRA_HEADERS"))
+
+        if extra_headers_json:
+            try:
+                parsed_headers = json.loads(extra_headers_json)
+                if isinstance(parsed_headers, dict):
+                    extra_headers = parsed_headers
```
Tomas2D, line 83: The parsing does not seem correct. In `TS` the input is not JSON, but a string with commas (KEY=VALUE,KEY2=VALUE2).

planetf1: Agreed, in that it was intentionally written to be a json string. As you point out this differs from typescript.

Headers can get complex, and need escaping if we want to handle the most general case, so going with a well defined format (like json) would seem to be more appropriate?

planetf1: it will also mean the env var will get set to the same value as the extra_headers property used by the ChatModel.

planetf1: In team call 20250312 consensus was that the extra headers should use simple strings, not json - will update

planetf1: * Updated implementation to use key value pairs.
* Tested with rits
* Error checking/parsing is simple, just checking for '='. Any issues will be met with a log entry, and ignoring the value set.
* No change to how extra_headers property is used. This remains in JSON format.

Tomas2D (commented):
If your project uses multiple LLM providers, then having just `LITELLM_EXTRA_HEADERS` would not work. Let's be provider-specific.

planetf1:
Updated implementation to:
 - use common util to manage headers
 - added unit test for utility
 - have a provider specific OPENAI_EXTRA_HEADERS
 - updated docs
 - tested against internal inference service
 
 This can be applied to the other litellm based adapters. I suggest we complete this PR first (and merge when approved). The others can be added in a subsequent PR

Tomas2D (approved):

Issue #531, opened 2025-03-07
Create ToolCallingAgent


Tomas2D:
Implement a Tool Calling Agent which either calls a tool or returns a final answer.

PR #551, opened 2025-03-10
feat(agents): add tool calling agent

Tomas2D:
```
python examples/agents/tool_calling.py
```

Ref: #531

### DIFF
--- a/python/beeai_framework/adapters/litellm/chat.py
+++ b/python/beeai_framework/adapters/litellm/chat.py
@@ -25,7 +25,7 @@ from litellm import (  # type: ignore
     get_supported_openai_params,
 )
 from litellm.types.utils import StreamingChoices
-from openai.lib._pydantic import to_strict_json_schema  # TODO: reimplement later on
+from openai.lib._pydantic import to_strict_json_schema  # TODO: look for an alternative or reimplement
 
 from beeai_framework.adapters.litellm._patch import _patch_litellm_cache
 from beeai_framework.backend.chat import (
@@ -161,6 +161,7 @@ class LiteLLMChatModel(ChatModel, ABC):
                     "function": {
                         "name": tool.name,
                         "description": tool.description,
+                        # OpenAI API requires more strict schema in order to perform well and pass the validation.
                         "parameters": to_strict_json_schema(tool.input_schema),
                     },
                 }
--- a/python/beeai_framework/adapters/litellm/chat.py
+++ b/python/beeai_framework/adapters/litellm/chat.py
@@ -25,6 +25,7 @@ from litellm import (  # type: ignore
     get_supported_openai_params,
 )
 from litellm.types.utils import StreamingChoices
+from openai.lib._pydantic import to_strict_json_schema  # TODO: reimplement later on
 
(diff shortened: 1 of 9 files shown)


python/beeai_framework/agents/base.py
```diff
@@ -14,56 +14,68 @@
...
-        execution: AgentExecutionConfig | None = None,
-        signal: AbortSignal | None = None,
+        self, input: ModelLike[TInput], options: ModelLike[TOptions] | None = None, /, **kwargs: Any
     ) -> Run[TOutput]:
         if self.is_running:
             raise RuntimeError("Agent is already running!")
 
+        # TODO: add error handling
+        input_formatted = to_model(self._schema.input, input)
+        options_formatted = to_model(self._schema.options, options or kwargs)
```
ajbozarth, line 58: This won't work as it does now since kwargs would contain the contents of both `input` and `options`. If we want to keep `run` flat we will need to either find a way to distinguish input from options, or we only flatten input and not options. I believe the following would do it:
```suggestion
        self, input: ModelLike[TInput], options: ModelLike[TOptions] | None = None, /, **kwargs: Any
    ) -> Run[TOutput]:
        if self.is_running:
            raise RuntimeError("Agent is already running!")

        # TODO: add error handling
        input_formatted = to_model(self._schema.input, input  or exclude_keys(kwargs, ["options"]))
        options_formatted = to_model(self._schema.options, options or kwargs["options"])
```

ajbozarth: I checked out your code locally to tinker a bit and it's actually even easier than what I put above. Give the flexibility of `to_model` we can actually just do the following and retain most of the current functionality. The only caveat would be we can't do `run("query string")` anymore, but change back to`run(prompt="query string")`
```suggestion
        self, input: ModelLike[TInput] | None = None, options: ModelLike[TOptions] | None = None, /, **kwargs: Any
    ) -> Run[TOutput]:
        if self.is_running:
            raise RuntimeError("Agent is already running!")

        # TODO: add error handling
        input_formatted = to_model(self._schema.input, input or kwargs)
        options_formatted = to_model(self._schema.options, options or kwargs)
```

Tomas2D: You are mixing two concepts - input/options.

With my approach, you must pass an input and all other arguments are interpreted as options.

ajbozarth: resolved by #562

python/beeai_framework/adapters/litellm/chat.py
```diff
@@ -141,7 +161,7 @@ def _transform_input(self, input: ChatModelInput) -> dict[str, Any]:
                     "function": {
                         "name": tool.name,
                         "description": tool.description,
-                        "parameters": tool.input_schema.model_json_schema(mode="validation"),
+                        "parameters": to_strict_json_schema(tool.input_schema),
```
Tomas2D, line 164: If the schema is more complex, it will not be processed by the OpenAI API. The `to_strict_json_schema` simplifies the shema.

ajbozarth (approved):

Issue #576, opened 2025-03-12
mypy - reporting errors on every commit


planetf1:
When trying to commit a code change, mypy is now reporting errors, even in files not changed:
```
(beeai-framework-py3.13) ➜  python git:(context) ✗ git commit -s -m 'chore(internal): Support context dictionary in FrameworkError'
Running Python pre-commit hook...
Poe <= hook
Poe => TYPE=${type} ./scripts/copyright.sh
Running 'nwa' version v0.5.2
2025/03/12 17:58:37 WARN file already has a header path=beeai_framework/__init__.py
2025/03/12 17:58:37 WARN file already has a header path=beeai_framework/adapters/__init__.py
2025/03/12 17:58:37 WARN file already has a header path=beeai_framework/adapters/amazon_bedrock/__init__.py
2025/03/12 17:58:37 WARN file already has a header path=beeai_framework/adapters/amazon_bedrock/backend/__init__.py
2025/03/12 17:58:37 WARN file already has a header path=beeai_framework/adapters/amazon_bedrock/backend/chat.py
2025/03/12 17:58:37 WARN file already has a header path=beeai_framework/adapters/anthropic/__init__.py
2025/03/12 17:58:37 WARN file already has a header path=beeai_framework/adapters/anthropic/backend/__init__.py
2025/03/12 17:58:37 WARN file already has a header path=beeai_framework/adapters/anthropic/backend/chat.py
2025/03/12 17:58:37 WARN file already has a header path=beeai_framework/adapters/azure_openai/__init__.py
2025/03/12 17:58:37 WARN file already has a header path=beeai_framework/adapters/azure_openai/backend/__init__.py
2025/03/12 17:58:37 WARN file already has a header path=beeai_framework/adapters/azure_openai/backend/chat.py
2025/03/12 17:58:37 WARN file already has a header path=beeai_framework/adapters/groq/__init__.py
2025/03/12 17:58:37 WARN file already has a header path=beeai_framework/adapters/groq/backend/__init__.py
2025/03/12 17:58:37 WARN file already has a header path=beeai_framework/adapters/groq/backend/chat.py
2025/03/12 17:58:37 WARN file already has a header path=beeai_framework/adapters/langchain/__init__.py
2025/03/12 17:58:37 WARN file already has a header path=beeai_framework/adapters/langchain/tools.py
2025/03/12 17:58:37 WARN file already has a header path=beeai_framework/adapters/litellm/_patch.py
[...]

PR #579, opened 2025-03-12
ci: add python e2e test workflow

ajbozarth:
This adds the following improvements:
- Add new `Python - E2E Tests` workflow that will run when a push to `main` includes changes in `/python/`
    - similar to the `TypeScript - E2E Examples` workflow, it installs Ollama to run tests
    - this can be extended later with env vars to also test examples using other model services like watsonx
- Document that `poetry install --all-extras` is usually needed for contributing developers
- Add `poe docs --type check` to pre-commit
- Add missing unit test decorators to `tests/backend/providers/test_azure_openai.py`


python/CONTRIBUTING.md
```diff
@@ -85,7 +85,7 @@ cd beeai-framework
 cd python
 
 # Use Poetry to install the project dependencies and activate a virtual environment
-poetry install
+poetry install # poetry install --all-extras
```
AngeloDanducci, line 88: Given the mypy problems we saw in #576 should all extras be on a separate line so it appears mandatory?

The way it currently is makes it appear optional to me.

ajbozarth: I did it this was since that was how we did `poe format --fix above` and not everyone following these step plan to contribute code, some just want a local build to play with the cutting edge. But I'm ok with just replacing `poetry install` with `poetry install --all-extras entirely here

Tomas2D (commented):
- Running examples (testing the whole flow) are E2E tests and not integration tests. 
- Running all of them for each PR is unsustainable (free tier).

If we want to really test some E2E tests (examples) then I would pick just a small subset of them.

ajbozarth:
So so based on review I have streamlined this PR and updated the top level description to match.

I've reverted all the e2e vs integration changes and remove e2e tests from the normal python workflow. 

Now e2e tests as a whole, including ones that need Ollama, are run in a separate workflow that only runs on `main`. This will enable discovering e2e tests failing after a push to `main` without running them all on every push to a PR.

As this is a less intrusive update we can probably merge as is

PR #849, opened 2025-05-26
feat: improve emitter interface

Tomas2D:

Closes: #

### Description

- allow to use `on` method as a decorator
- add `off` method for custom cleanups
- update examples
- add tests
- make `match` method deprecated as it was here for the parity with TS where it make sense becase of typings

#### General
- [ ] I have read the appropriate contributor guide: [Python](https://github.com/i-am-bee/beeai-framework/blob/main/CONTRIBUTING.md)
- [ ] I have signed off on my commit: [Python instructions](https://github.com/i-am-bee/beeai-framework/blob/main/python/CONTRIBUTING.md#developer-certificate-of-origin-dco) / [TypeScript instructions](https://github.com/i-am-bee/beeai-framework/blob/main/typescript/CONTRIBUTING.md#developer-certificate-of-origin-dco)
- [ ] Commit messages and PR title follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary)

#### Code quality checks
- [ ] Linting passes: Python `poe lint` or `poe lint --fix` / TypeScript `yarn lint` or `yarn lint:fix`
- [ ] Formatting is applied: Python `poe format` or `poe format --fix` / TypeScript: `yarn format` or `yarn format:fix`
- [ ] Static type checks pass: Python `poe type-check`

#### Testing
- [ ] Unit tests pass: Python `poe test --type unit` / TypeScript `yarn test:unit`
- [ ] E2E tests pass: Python `poe test --type e2e` / TypeScript: `yarn test:e2e`
- [ ] Integration tests pass: Python `poe test --type integration`

#### Documentation
- [ ] Embedme embeds code examples in docs. To update after edits, run: Python `poe docs --type build`

### DIFF
--- a/docs/integrations/a2a.mdx
+++ b/docs/integrations/a2a.mdx
@@ -98,7 +98,9 @@ A2AServer lets you expose agents built in the BeeAI framework via A2A protocol.
 	    # For the ToolCallingAgent, we don't need to specify ACPAgent factory method
 	    # because it is already registered in the A2AServer
 	    # we use LRU memory manager to keep limited amount of sessions in the memory
-	    A2AServer(config=A2AServerConfig(port=9999), memory_manager=LRUMemoryManager(maxsize=100)).register(agent).serve()
+	    A2AServer(
+	        config=A2AServerConfig(port=9999, protocol="jsonrpc"), memory_manager=LRUMemoryManager(maxsize=100)
+	    ).register(agent).serve()
 	
 	
 	if __name__ == "__main__":
--- a/docs/modules/serve.mdx
+++ b/docs/modules/serve.mdx
@@ -51,7 +51,9 @@ def main() -> None:
     # For the ToolCallingAgent, we don't need to specify ACPAgent factory method
     # because it is already registered in the A2AServer
     # we use LRU memory manager to keep limited amount of sessions in the memory
-    A2AServer(config=A2AServerConfig(port=9999), memory_manager=LRUMemoryManager(maxsize=100)).register(agent).serve()
+    A2AServer(
+        config=A2AServerConfig(port=9999, protocol="jsonrpc"), memory_manager=LRUMemoryManager(maxsize=100)
+    ).register(agent).serve()
 
 
 if __name__ == "__main__":
--- a/python/beeai_framework/adapters/a2a/agents/_utils.py
+++ b/python/beeai_framework/adapters/a2a/agents/_utils.py
@@ -47,22 +47,3 @@ def convert_a2a_to_framework_message(input: a2a_types.Message | a2a_types.Artifa
                 )
             )
     return msg
-
-
-def has_content(event: a2a_types.SendStreamingMessageResponse) -> bool:
-    """Check if the event has content."""
-    if isinstance(event.root, a2a_types.SendStreamingMessageSuccessResponse):
-        response = event.root.result
-        if isinstance(response, a2a_types.Message):
-            return True
-        elif isinstance(response, a2a_types.TaskArtifactUpdateEvent):
-            return response.last_chunk or False
-        elif isinstance(response, a2a_types.TaskStatusUpdateEvent):
-            return bool(response.status.message)
-        elif isinstance(response, a2a_types.Task):
-            return bool(response.status.message) or bool(response.artifacts and len(response.artifacts) > 0)
-        else:
-            return False
-    elif isinstance(event.root, a2a_types.JSONRPCErrorResponse):
-        return True
-    return False
--- a/python/beeai_framework/adapters/a2a/agents/agent.py
+++ b/python/beeai_framework/adapters/a2a/agents/agent.py
@@ -1,18 +1,21 @@
 # Copyright 2025 © BeeAI a Series of LF Projects, LLC
 # SPDX-License-Identifier: Apache-2.0
 
-from collections.abc import AsyncGenerator
+from collections.abc import Sequence
 from typing import Unpack
 from uuid import uuid4
 
 import httpx
+from a2a.types import TransportProtocol
 
-from beeai_framework.adapters.a2a.agents._utils import convert_a2a_to_framework_message, has_content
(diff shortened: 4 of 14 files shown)


xjacka (approved):

python/beeai_framework/emitter/emitter.py
```diff
@@ -126,53 +127,102 @@ def destroy(self) -> None:
...
+        if callback is None or event is None:
+
+            def decorator(fn: Callback) -> Callback:
+                name = event or str(fn.__name__).removeprefix("on_").removeprefix("handle_")
+                self._register(name, fn, options)
+                return fn
+
+            return decorator
+        else:
+            return self._register(event, callback, options)
```
araujof: The overloads provide a nice developer experience with type hints, but the interface complexity suggests it might benefit from some revisions. A few considerations below, as I tried to understand the implementation.

**Why define it this way**

This interface uses `@overload` decorators to provide multiple type signatures for a single `on` method.

This supports two usage patterns: Decorator and Direct Call Pattern

```python
# As a decorator (when callback=None)
@emitter.on('event')
def handler():
    pass

# As a direct call
cleanup = emitter.on('event', handler)
```

The return types:
- When `callback=None` → returns a decorator function
- When `callback` is provided → returns a cleanup function

**Potential Anti-patterns and Suggestions for Improvement**

1. The `event` parameter can be `None` in the first overload but must be present in the second. This asymmetry is confusing, and error-prone. See the issue raised by Gemini. How about we make this contract consistent? 
- [x] Make `event` parameter required in all forms of this interface
 
2. Perhaps, a simpler alternative can be built along the lines of the following excerpt: 
```python
def on(self, event: Matcher, callback: Callback | None = None) -> CleanupFn | Callable:
    if callback is None:
        return lambda cb: self.on(event, cb)
    # register and return cleanup
``````

3. Despite the overloads, the actual implementation still returns a union type, which somewhat defeats the purpose of the overloads for runtime behavior. This may suggest that splitting decorator from direct call implementation is a simpler approach (at the cost of a less user friendly approach where `on` can be used both ways). Personally, I'm okay with keeping it this way, as long we document things a bit better. So, how about:
- [ ] Update the `on` method’s docstring to explicitly describe both usage patterns—direct-call vs decorator, and what each returns.

4. `None` values for `event` and `callback` have special meaning rather than just being absent, which can be error-prone. This can be addressed by taking into consideration point 1 above (which corroborates with Gemini's review).

File: python/beeai_framework/emitter/emitter.py [hunk 2]
```diff
@@ -126,53 +127,102 @@ def destroy(self) -> None:
...
+
+        for listener in list(self._listeners):
+            if callback is not None and listener.callback is not callback:
+                continue
+            if event is not None and listener.raw != event:
+                continue
+            if options is not None and listener.options != options:
+                continue
+
+            self._listeners.remove(listener)
```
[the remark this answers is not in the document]
araujof: I think I agree with Gemini on this ;) This comparison breaks for regex patterns (`re.Pattern`) and callable matchers because two logically equivalent regexes or lambdas won't match via `!=`. Because the Matches can be many things (`str`, `re.Pattern[str]`, `Callable`), I think we may need a more robust matcher for the events. 

**Suggestion**
```python
if isinstance(event, re.Pattern) and isinstance(listener.raw, re.Pattern):
    if event.pattern != listener.raw.pattern or event.flags != listener.raw.flags:
        continue
elif callable(event) and callable(listener.raw):
    # Possibly compare function identities or wrap them to allow matching.
    if event is not listener.raw:
        continue
else:
    if listener.raw != event:
        continue
```

If you think this is reasonable, consider adding a `matcher_equal()` helper to encapsulate these checks cleanly.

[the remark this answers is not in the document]
araujof: Consider adding some unit tests to cover deregistration for regex or callable matchers of the `off` method.

Examples: 
```python
# Regex matcher deregistration
regex = re.compile(r"^a.*$")
emitter.on(regex, handler)
emitter.off(regex)
await emitter.emit("abc", None)
assert handler_not_called

# Callable matcher deregistration
predicate = lambda event: event.path.endswith("done")
emitter.on(predicate, handler)
emitter.off(predicate)
await emitter.emit("task.done", None)
assert handler_not_called
```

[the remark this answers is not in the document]
Tomas2D: Good catch 👍🏻 . Updated.

python/beeai_framework/emitter/emitter.py
Tomas2D: I wanted to have two separate functions. One for general registration and one for the decorator. This would nicely simplify the code, but could hurt the usage experience.

I removed the check whether `event` is None or not. Now the function returns either `Callable[[Callback], Callback]` or a `CleanupFn`.

The follow-up question might be, why not just return the cleanup function in both cases? 

```python
@emitter.on("new_token")
async def handle_new_token(data: Any, __: Any) -> None:
  pass

# the function is now cleanup and the usage is very cubersome
# the signature is confusing to the IDE (it was async, now it is not)
# the name is misleading, instead of calling the function, you are calling a cleanup
handle_new_token() 
```

araujof: Agree. I commented on this above. I think we can keep the overloads. Shouldn't `event` be positional and required in the method signatures? Is there a case for not specifying an `event` (as decorator and direct call)?

araujof (approved):
LGTM

- [ ] Drop the `print(f*)` statement
- [ ] Consider making `event` required in `on` method (I think it's still optional, maybe I misread it, or there is a case for them to be optional; if so, ignore this comment)

python/beeai_framework/emitter/emitter.py
Tomas2D: An event must be specified for a direct call. It is optional for the decorator, as it is retrieved from the function name if not specified.

```
@emitter.on()
async def handle_new_token(data: Any, __: Any) -> None:
  pass
```

In this case, the event name will be `new_token`. If you think it is not useful, we can remove it.

PR #974, opened 2025-08-05
feat(backend): Adding text splitter backend class

antonpibm:
This feature implements the text splitter abstraction for RAG and provides a LangChain integration

This brings us to the point where there are no external imports in the rag agent example

### DIFF
--- a/python/beeai_framework/adapters/langchain/backend/text_splitter.py
+++ b/python/beeai_framework/adapters/langchain/backend/text_splitter.py
@@ -1,0 +1,78 @@
+# Copyright 2025 © BeeAI a Series of LF Projects, LLC
+# SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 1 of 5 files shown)


python/beeai_framework/adapters/langchain/backend/text_splitter.py
```diff
@@ -0,0 +1,65 @@
...
+from beeai_framework.backend.types import Document
+from beeai_framework.logger import Logger
+
+logger = Logger(__name__)
+
+
+class LangChainTextSplitter(TextSplitter):
+    def __init__(self, *, text_splitter: LCTextSplitter) -> None:
+        super().__init__()
+        self.text_splitter: LCTextSplitter = text_splitter
```
Tomas2D, line 27: explicit type should not be needed as it is auto-inferred

Issue #938, opened 2025-07-21
Introduce Runnable interface to unify “runnables”


jenna-winkler:
## Problem Statement

BeeAI Framework currently implements many objects that can be invoked with a `run` method, but these implementations are inconsistent and lack a unified interface. This creates several issues:

1. **Inconsistent Method Signatures**: The arguments, argument names, and argument types of `run` methods vary across objects, even for objects of the same kind
2. **Lack of Type Safety**: Many classes use `*args: Any, **kwargs: Any` patterns that provide flexibility but sacrifice:
   - Explicitness (unclear function signatures)
   - Type checking capabilities
   - Documentation clarity
3. **Loosely Defined Interfaces**: For example, `BaseAgent` currently defines:
   ```python
   @abstractmethod
   def run(self, *args: Any, **kwargs: Any) -> Run[TOutput]:
       pass
   ```

This approach limits discoverability, makes the codebase harder to maintain, and places a burden on developers to learn and manage heterogeneous abstractions.

## Proposed Solution

Introduce a new `Runnable` interface that defines explicit types for `input`, `output`, and execution `options` for runnable objects.

**Key Points:**

- **Unified `run` Method:** Standardizes the `run` method signature across modules.
- **Consistent Input Handling:** The `input` parameter is always the first and only positional argument.
- **No Generics:** Avoids generics in favor of standardizing input/output types within implementations.
- **Type Hints Preserved:** Maintains type hints for better developer experience and static analysis.
- **Simplified Implementation:** Users can leverage decorators instead of manual invocation of `RunContext.enter()` and handler wrapping.
- **Bonus** it does not drastically break existing interfaces, easy to migrate.

### Core Interface Design

```python
class RunnableOptions(TypedDict, total=False):
    signal: AbortSignal
    """The runnable's abort signal data"""
    context: dict[str, Any]
    """Context can be used to pass additional context to runnable"""
    # It can be subclassed to include additional options

class RunnableOutput(BaseModel):
    output: list[AnyMessage]
    """The runnable output"""
    context: dict[str, Any] | None = None
    """Context can be used to return additional data by runnable"""
    # It can be subclassed to add additional fields as needed

class Runnable(ABC):
    @abstractmethod
    def run(self, input: list[AnyMessage], /, **kwargs: Unpack[RunnableOptions]) -> Run[RunnableOutput]: ...

    @property
    @abstractmethod
    def emitter(self) -> Emitter: ...
```

 **Notes:**
[...]

PR #982, opened 2025-08-06
feat: introduce runnable interface

araujof:

#938 

### Description

This PR introduces a `Runnable` class to unify runnables in the framework.

Notes:
- Execution middleware is included as an instance attribute of runnables, such that the `runnable_entry` decorator can automatically inject the middleware during wrapping. The runnable constructor defaults its value to an empty list.

#### General

#### Code quality checks

#### Testing
- [x] E2E tests pass: `mise test:e2e`

#### Documentation
- [ ] Embedme embeds code examples in docs. To update after edits, run: Python `mise docs:fix`

### DIFF
--- a/python/beeai_framework/runnable.py
+++ b/python/beeai_framework/runnable.py
@@ -6,7 +6,7 @@ from abc import ABC, abstractmethod
 from collections.abc import Awaitable, Callable
 from typing import Any, TypedDict, Unpack
 
-from pydantic import BaseModel
+from pydantic import BaseModel, ConfigDict
 from typing_extensions import ParamSpec, TypeVar
 
 from beeai_framework.backend import AnyMessage
@@ -31,6 +31,8 @@ class RunnableOptions(TypedDict, total=False):
 
 class RunnableOutput(BaseModel):
     """Runnable output."""
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
 
     output: list[AnyMessage]
     """The runnable output"""
--- a/mise.toml
+++ b/mise.toml
@@ -123,6 +123,6 @@ depends = ["common:check"]
 [tasks."common:git-hooks:commit-msg"]
 dir = "{{config_root}}"
 run = """
-grep -qE "^(fixup! )*(feat|fix|chore|ci|docs|revert)(\\(\\w+\\))?!?: .+" .git/COMMIT_EDITMSG || (echo "ERROR: Commit message does not adhere to conventional commit style!" && exit 1)
+grep -qE "^(fixup! )*(feat|fix|chore|ci|docs|revert|refactor|test)(\\(\\w+\\))?!?: .+" .git/COMMIT_EDITMSG || (echo "ERROR: Commit message does not adhere to conventional commit style!" && exit 1)
 grep -qi "^Signed-off-by: "                                    .git/COMMIT_EDITMSG || (echo "ERROR: Commit message must contain a Signed-off-by line!"            && exit 1)
 """
--- a/python/tests/test_runnable.py
+++ b/python/tests/test_runnable.py
@@ -1,0 +1,67 @@
+# Copyright 2025 © BeeAI a Series of LF Projects, LLC
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import Any, Unpack
+
+import pytest
+
+from beeai_framework.backend import AnyMessage, AssistantMessage, UserMessage
+from beeai_framework.context import RunContext, RunMiddlewareType
+from beeai_framework.emitter import Emitter, EventMeta
+from beeai_framework.runnable import Runnable, RunnableOptions, RunnableOutput, runnable_entry
+
+"""
+Utility functions and classes
+"""
+
+
+class MyRunnable(Runnable):
+    def __init__(self, middlewares: list[RunMiddlewareType] | None = None) -> None:
+        super().__init__(middlewares)
+
+    @property
+    def emitter(self) -> Emitter:
+        return Emitter.root().child(namespace=["runnable", "main"])
+
+    @runnable_entry
+    async def run(self, input: list[AnyMessage], /, **kwargs: Unpack[RunnableOptions]) -> RunnableOutput:
+        ctx = RunContext.get()
+        await ctx.emitter.emit("run", f"input: {input[-1].text}, context: {kwargs.get('context')}")
+        return RunnableOutput(output=[AssistantMessage(content="Hi, there!")])
+
+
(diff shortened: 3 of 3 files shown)


matoushavlena (approved):

python/beeai_framework/runnable.py
```diff
@@ -0,0 +1,109 @@
...
+            # ... agent logic ...
+            return AgentOutput(output=...)
+
+    Attributes:
+        _middleware: The middleware to be used when executing the runnable.
+    """
+
+    def __init__(self, middleware: list[RunMiddlewareType] | None = None) -> None:
+        super().__init__()
+        self._middleware = middleware or []
```
Tomas2D: I would personally go with a plural.

python/beeai_framework/runnable.py
```diff
@@ -0,0 +1,109 @@
...
+    For convenience, concrete subclasses can decorate `run` with `runnable_entry`
+    to automatically wrap the runnable into an execution context:
+
+        @runnable_entry
+        async def run(self, input: str | list[AnyMessage], /, **kwargs: Unpack[AgentOptions]) -> RunnableOutput:
+            # ... agent logic ...
+            return AgentOutput(output=...)
+
+    Attributes:
+        _middleware: The middleware to be used when executing the runnable.
```
Tomas2D: The `middleware` attribute in the constructor is without the prefix `_`.

python/beeai_framework/runnable.py
```diff
@@ -0,0 +1,109 @@
...
+        if not isinstance(self, Runnable):
+            raise TypeError("The first argument of a runnable must be a Runnable instance.")
+
+        runnable_kwargs: RunnableOptions = kwargs  # type: ignore
+        return RunContext.enter(
+            self,
+            inner,
+            signal=runnable_kwargs.get("signal", None),
+            run_params={"input": args[1], **exclude_keys(kwargs, {"signal", "input"})},
+        ).middleware(*self.middleware)
```
Tomas2D: Would like to propagate the `context` from kwargs to the RunContext? If so, then we could do `.middleware(*self.middleware).context(kwargs.get("context"))`.

python/beeai_framework/runnable.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class Runnable(ABC):
+    """A unit of work that can be invoked using a stable interface.
+
+    For convenience, concrete subclasses can decorate `run` with `runnable_entry`
+    to automatically wrap the runnable into an execution context:
+
+        @runnable_entry
+        async def run(self, input: str | list[AnyMessage], /, **kwargs: Unpack[AgentOptions]) -> RunnableOutput:
```
antonpibm: AgentOptions is not defined, I suggest to completely remove this comment and mention this in the documentation

python/beeai_framework/runnable.py
araujof: Updated name to plural.

python/beeai_framework/runnable.py
araujof: The docstring `Attributes:` has a special meaning here, to document class attributes (not properties or constructor arguments). I wanted to somehow make it clear that runnables have this instance attribute. This is just for API documentation.

python/beeai_framework/runnable.py
araujof: Updated

araujof:
@Tomas2D Thanks for bringing this up! The getter for the `RunContext` was missing. I will fix that.

> The `runnable_entry` does not propagate the context (which is based on my finding, even not possible if we want to preserve typing). However, the creator needs access to it to emit events, etc.
> 
> ```python
> @runnable_entry
> def run(self, input: list[AnyMessage], /, **kwargs: Unpack[RunnableOptions]) -> Run[RunnableOutput]:
>     ctx = RunContext.get() # this needs to be implemented
> ```
> 
> The implementation (extension of the `RunContext` class) then would be
> 
> ```python
> class RunContext:
>     ...
> 
>     @classmethod
>     def get(cls) -> "RunContext":
>         context = storage.get(None)
>         If context is None:
>             raise RuntimeError("Called from non-context.")
>         return context
> ```

A clarification on the question about propagating `RunOptions.context`:
> Would like to propagate the context from kwargs to the RunContext? If so, then we could do .middleware(*self.middleware).context(kwargs.get("context")).

We already pass the kwargs to the runnable's `run` method when passing the `run_params` to the `RunContext.enter()`:
> run_params={"input": args[1], **exclude_keys(kwargs, {"signal", "input"})},

1. Are there any specific use cases or context parameters that need to be passed to RunContext through `run_params`? If so, maybe we could disambiguate and make them first-class options, like `signal`, in `RunnableOptions`.  What do you think?
2. Is there a reason to pass `signal` as a kwarg to `run()`, or should we move it as a class instance attribute (e.g., `self._signal`, like we have for middlewares), in which case we could define a default, and allow concrete classes to optionally pass a custom signal? If we do this, we can pass the signal to `RunContext.enter()` from the runnable instance `self._signal`.

Tomas2D:
> We already pass the kwargs to the runnable's run method when passing the run_params to the RunContext.enter():

Yes, but we differentiate between `context` as a kwarg and `Run`'s context. Now you are passing the `context` (to the `run_params` - useful for logging etc.), but you are not setting the `Run`s `context` field, which might be okay if we want to distinguish between those too. Run's context is part of the telemetry etc, and can be accessed in nested calls.

See how it is used in the `HandoffTool` [link](https://github.com/i-am-bee/beeai-framework/blob/7e9d031eb5451e28e49ba75ba6ee864552710e84/python/beeai_framework/tools/handoff.py#L42) and how the context is set https://github.com/i-am-bee/beeai-framework/blob/7e9d031eb5451e28e49ba75ba6ee864552710e84/python/beeai_framework/agents/tool_calling/agent.py#L214 (example for `ToolCallingAgent`)

>Are there any specific use cases or context parameters that need to be passed to RunContext through run_params? If so, maybe we could disambiguate and make them first-class options, like signal, in RunnableOptions. Is there a reason to pass signal as a kwarg to run(), or should we move it as a class instance attribute (e.g., self._signal, like we have for middlewares), in which case we could define a default, and allow concrete classes to optionally pass a custom signal? If we do this, we can pass the signal to RunContext.enter() from the runnable instance self._signal.

Signals are bound to the Run's object and therefore can't be attached to a class because the user can call the run method simultaneously. In addition, the signal can nest in the call stack.

araujof:
@Tomas2D @jenna-winkler - 

I addressed all comments and change requests and added unit tests. 
Please give it a final check and merge.

Tomas2D (approved):
All set. Great job! 🚀

PR #991, opened 2025-08-07
feat(backend): Vector store tool

antonpibm:
This PR adds support for vector stores as a tool, together with a usage example as part of RequriementsAgent


python/examples/agents/experimental/requirement/rag.py
```diff
@@ -0,0 +1,121 @@
...
+from beeai_framework.backend.embedding import EmbeddingModel
+from beeai_framework.backend.text_splitter import TextSplitter
+from beeai_framework.backend.vector_store import VectorStore
+from beeai_framework.logger import Logger
+from beeai_framework.memory import UnconstrainedMemory
+from beeai_framework.middleware.trajectory import GlobalTrajectoryMiddleware
+from beeai_framework.tools import Tool
+from beeai_framework.tools.search.retrieval import VectorStoreSearchTool
+
+logger = Logger("vector-store-search-example", level=logging.DEBUG)
```
Tomas2D: You are initiating a logger class, but you are not using it anywhere.

python/examples/agents/experimental/requirement/rag.py
```diff
@@ -0,0 +1,121 @@
...
+
+    # Alternative: Create search tool using dynamic loading
+    # embedding_model = EmbeddingModel.from_name("watsonx:ibm/slate-125m-english-rtrvr-v2", truncate_input_tokens=500)
+    # search_tool = VectorStoreSearchTool.from_name(
+    #     name="beeai:TemporalVectorStore",
+    #     embedding_model=embedding_model
+    # )
+
+    # Create RequirementAgent with the vector store search tool
+    llm = ChatModel.from_name("ollama:llama3.2")
```
Tomas2D: In all examples for RequirementAgent we use either `llama3.1:8b` or `granite3.3:8b`. Could you update the model please?

python/beeai_framework/tools/search/base.py
```diff
@@ -11,7 +11,7 @@
 
 class SearchToolResult(BaseModel):
     title: str
-    description: str
+    content: str
```
Tomas2D: This would be a breaking change. I think that you don't need to inherit from that class.

Tomas2D: I would create a custom class for the RAG Tool outputs (`RAGToolOutput`).

antonpibm: As this is only the retrieval part of RAG, and I believe it's better suited as a search tool, reverting this change

PR #1004, opened 2025-08-14
feat(serve): switch BeeAIPlatform agent and server to beeai-sdk and a2a

xjacka:
### Description

Migrate BeeAIPlatformServer a BeeAIPlatformAgent to beeai-sdk and a2a

#### General
- [x] I have read the appropriate contributor guide: [Python](https://github.com/i-am-bee/beeai-framework/blob/main/CONTRIBUTING.md)
/ [TypeScript](https://github.com/i-am-bee/beeai-framework/blob/main/CONTRIBUTING.md)
- [x] I have signed off on my commit: [Python instructions](https://github.com/i-am-bee/beeai-framework/blob/main/python/CONTRIBUTING.md#developer-certificate-of-origin-dco) / [TypeScript instructions](https://github.com/i-am-bee/beeai-framework/blob/main/typescript/CONTRIBUTING.md#developer-certificate-of-origin-dco)
- [x] Commit messages and PR title follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary)
- [x] Appropriate label(s) added to PR: `Python` for Python changes, `TypeScript` for TypeScript changes

#### Code quality checks
- [x] Code quality checks pass: `mise check` (`mise fix` to auto-fix)

#### Testing
- [x] Unit tests pass: `mise test:unit`
- [ ] E2E tests pass: `mise test:e2e`
- [ ] Tests are included (for bug fixes or new features)

#### Documentation
- [x] Documentation is updated
- [x] Embedme embeds code examples in docs. To update after edits, run: Python `mise docs:fix`

PR #1013, opened 2025-08-20
refactor!: convert agents to runnables

### LINKED ISSUE #938: Introduce Runnable interface to unify “runnables”

araujof:
## Overview

This PR implements a major refactoring of the BeeAI Framework to convert all agents into `Runnable`'s, establishing a consistent and unified interface that enables composition and interoperability with other framework components.

## Related Issues

Closes #938

## Background

Previously, BeeAI Framework implemented many objects with `run` methods, but these implementations were inconsistent and lacked a unified interface. This created several issues:

- **Inconsistent Method Signatures**: Arguments, names, and types varied across objects
- **Lack of Type Safety**: Many classes used `*args: Any, **kwargs: Any` patterns
- **Loosely Defined Interfaces**: Limited discoverability and maintainability

### Core Interface Implementation

- **Standardized `run` Method**: Unified signature across all agent objects
- **Normalized types around a `BaseAgent` interface**: Defines explicit types for `input`, `output`, and execution `options` for agents
- **Make `RunnableOutput` a Base Type**: Common output structure with `output` and optional `context` fields
- **Created `AgentOutput` Type**: Extends `RunnableOutput` with `structured_output` field for agent-specific needs

### Agent Refactoring

- **Updated `BaseAgent`**: Now implements the `Runnable` interface
- **Enhanced Type Safety**: Parametric types on `Runnable` and `BaseAgent` with proper bounds
- **Backward Compatibility**: Agent output types can be specialized with custom fields for existing functionality
- **Flexible Input Types**: Agents accept both `str` and `list[AnyMessage]` for user convenience

### Key Design Decisions

1. **`AgentOutput` as Base Type**: Subclasses `RunnableOutput` to include structured output field
2. **Backward Compatibility**: Specialized output types maintain custom fields during migration period
3. **Parametric Types**: `Runnable` and `BaseAgent` are parametric on return types with proper defaults and bounds
4. **Type Safety**: Return types of base runnables and agents bounded to `RunnableOutput` and `AgentOutput` respectively

#### Examples

Let's define two agents and check a few usage examples.

First, the required imports.

```Python
from typing import Unpack
from beeai_framework.backend import AnyMessage
from beeai_framework.runnable import Runnable, RunnableOutput, RunnableOptions, runnable_entry
from beeai_framework.agents import BaseAgent, AgentOptions, AgentOutput
```

Note that from [agents/base.py](https://github.com/i-am-bee/beeai-framework/blob/e705a1b1a6f786c174ff928d0cca5f40b5cc572a/python/beeai_framework/agents/base.py#L52-L59), we have:

```Python
class AgentOutput(RunnableOutput):
    output_structured: Any | BaseModel = None
    """The formatted output returned by the agent."""

R = TypeVar("R", default=AgentOutput, bound=AgentOutput)

class BaseAgent(Runnable[R]):

    @abstractmethod
    def run(self, input: str | list[AnyMessage], /, **kwargs: Unpack[AgentOptions]) -> Run[R]:
```

Therefore, the following agent's return type defaults to `AgentOutput` type (no generic parameter needed in the class definition).

```Python
class MyAgent(BaseAgent):

    @runnable_entry
    async def run(self, input: str | list[AnyMessage], /, **kwargs: Unpack[AgentOptions]) -> AgentOutput: ...
```

Now, let's define a new output type, subclassing `AgentOutput` (typecheker will complain if it doesn't - see bound property in the `R` type above).

```Python
class AnotherAgentOutput(AgentOutput):
    state: dict

class AnotherAgent(BaseAgent[AnotherAgentOutput]): # here we specify the parametric return type

    @runnable_entry
    async def run(self, input: str | list[AnyMessage], /, **kwargs: Unpack[AgentOptions]) -> AnotherAgentOutput: ...
```

This allows us to refine the return type without violating type covariance. 

Some usage examples: 

```Python
async def main():

    myagent = MyAgent(...)
    result = await myagent.run("Hi", expected_output="format the output in markdown notation")

    # All properties of AgentOutput are here too
    print(result.output) # prints the list of messages in the output
    print(result.message.text) # prints the last message in output
    print(result.context) # prints the context
    print(result.output_structured) # prints the formatted output

    another = AnotherAgent(...)
    result = await another.run("Hi", expected_output="format the output in markdown notation")

    # All properties of AgentOutput are here too
    print(result.output) # prints the list of messages in the output
    print(result.message.text) # prints the last message in output
    print(result.context) # prints the context
    print(result.output_structured) # prints the formatted output

    # We can also access the state field from the agent output object
    print(result.state)
```

Without parametric `R`, all abstract methods for Runnable would look like this:

```Python
@runnable_entry
async def run(self, input: str | list[AnyMessage], /, **kwargs: Unpack[AgentOptions]) -> RunnableOutput: ...
```
And concrete usage would look like this:
```Python
another = AnotherAgent(...)
result = await another.run("Hi", expected_output="format the output in markdown notation")
print(result.state) # type checking error! We would need to convert/recast the output type every time
```

Now, the advantage is that we can still operate at the 'Runnable' abstraction level due to the inherit class hierarchy on the runnable output type:

```Python
async def run_text(runnable: Runnable, input: str, **kwargs: Unpack[RunnableOptions]) -> RunnableOutput:
    return await runnable.run(input, **kwargs)
```

## Benefits

- **Unified Developer Experience**: Learning one runnable/agent teaches you all others
- **Enhanced Type Safety**: Better IDE support, static analysis, and runtime validation  
- **Improved Composability**: Unified interface enables easier composition of different runnable types
- **Stable Semantics**: Consistent argument types and meanings across all runnable objects
- **Maintained Compatibility**: Existing code continues to work during migration

## Future Work

The following improvements can be tackled in future iterations:

### Agent Unification
- **Constructor Normalization**: Standardize agent constructors and initialization logic
- **Memory Handling**: Revisit memory update logic for unified agents
  - Auto-append new inputs to memory when configured
  - Consider implementation via run context, mixins, or decorators

### Code Simplification  
- **State Abstractions**: Simplify and deduplicate agent state management
- **Memory Management**: Unified approach to agent memory handling
- **Metadata Handling**: Streamline agent metadata abstractions

## Testing

- [x] All existing agent tests pass
- [x] New runnable interface tests added
- [x] Backward compatibility verified
- [x] Type checking validated

## Breaking Changes

This is a **breaking change** (hence the `!` in the commit message) as it modifies core agent interfaces. However, backward compatibility has been maintained where possible to ease migration.

#### General

#### Code quality checks

#### Testing
- [x] E2E tests pass: `mise test:e2e`

#### Documentation
- [x] Embedme embeds code examples in docs. To update after edits, run: Python `mise docs:fix`


python/beeai_framework/adapters/a2a/agents/agent.py
```diff
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from collections.abc import AsyncGenerator
+from typing import Unpack
```
Tomas2D, line 5: `Unpack` needs to be imported from `typing_extensions` to work in Python before 3.13.
This should be caught by Ruff or Mypy. Would you mind taking a look to see if there is a rule that can check it?

python/beeai_framework/adapters/a2a/agents/__init__.py
```diff
@@ -3,11 +3,11 @@
...
 from beeai_framework.adapters.a2a.agents.agent import A2AAgent
 from beeai_framework.adapters.a2a.agents.events import A2AAgentErrorEvent, A2AAgentUpdateEvent
-from beeai_framework.adapters.a2a.agents.types import A2AAgentRunOutput
+from beeai_framework.adapters.a2a.agents.types import A2AAgentOutput
 
 __all__ = [
     "A2AAgent",
     "A2AAgentErrorEvent",
-    "A2AAgentRunOutput",
+    "A2AAgentOutput",
```
Tomas2D, line 12: This is a breaking change. Perhaps I would re-export `A2AAgentRunOutput` as an alias to `A2AAgentOutput` to prevent the current applications from breaking.

python/beeai_framework/adapters/a2a/agents/agent.py
```diff
@@ -205,18 +199,22 @@ async def clone(self) -> "A2AAgent":
...
     def _convert_message_to_framework_message(
-        self, input: str | AnyMessage | a2a_types.Message | a2a_types.Artifact
+        self, input: str | list[AnyMessage] | AnyMessage | a2a_types.Message | a2a_types.Artifact
     ) -> AnyMessage:
         if isinstance(input, str):
             return UserMessage(input)
         elif isinstance(input, Message):
             return input
+        elif isinstance(input, list) and input and isinstance(input[-1], Message):
+            return input[-1]
```
Tomas2D: The only place where we are passing multiple messages is here.

https://github.com/i-am-bee/beeai-framework/blob/82120b5ed41f8eea26ed89dc5e34d3ff496e21f5/python/beeai_framework/adapters/a2a/agents/agent.py#L132-L134

If we used your changes, then we would lose all new messages except the last one.

In my opinion, we should keep the function as is and make this change instead.

```python
input_messages = [msg for msg in self._convert_message_to_framework_message(input)]
await self.memory.add_many(input_messages)
```

python/beeai_framework/adapters/a2a/agents/agent.py
```diff
@@ -205,18 +199,22 @@ async def clone(self) -> "A2AAgent":
...
         elif isinstance(input, a2a_types.Message | a2a_types.Artifact):
             return convert_a2a_to_framework_message(input)
         else:
             raise ValueError(f"Unsupported input type {type(input)}")
 
-    def _convert_to_a2a_message(self, input: str | AnyMessage | a2a_types.Message) -> a2a_types.Message:
+    def _convert_to_a2a_message(
```
Tomas2D, line 325: Not sure about this one. 

Your thoughts? @xjacka as I know you just finished #1006

python/beeai_framework/adapters/beeai_platform/agents/agent.py
```diff
@@ -1,5 +1,6 @@
 # Copyright 2025 © BeeAI a Series of LF Projects, LLC
 # SPDX-License-Identifier: Apache-2.0
+from typing import Unpack
```
Tomas2D, line 4:
```python
from `typing_extensions` import Unpack
```

+ I think that there should be a new line.

python/beeai_framework/adapters/watsonx_orchestrate/agents/agent.py
```diff
@@ -4,7 +4,7 @@
 import uuid
 from collections.abc import AsyncGenerator
 from contextlib import asynccontextmanager
-from typing import Literal
+from typing import Literal, Unpack
```
Tomas2D, line 7: same

python/beeai_framework/agents/base.py
```diff
@@ -1,28 +1,85 @@
...
+
+    max_iterations: int
+    """
+    Maximum number of iterations.
+    """
+
+    backstory: str
+    """
+    Additional piece of information or background for the agent.
+    """
```
Tomas2D: I know we kind of agreed on this offline because we don’t want to overuse the term *context*.  

Some frameworks, like `CrewAI`, allow users to set a `backstory` when constructing an agent. We currently use `instructions` instead. We should consider adding `backstory` to the constructor and using something else for the agent’s run.

python/beeai_framework/agents/base.py
```diff
@@ -1,28 +1,85 @@
 # Copyright 2025 © BeeAI a Series of LF Projects, LLC
 # SPDX-License-Identifier: Apache-2.0
 
-from abc import ABC, abstractmethod
+from abc import abstractmethod
 from collections.abc import Awaitable, Callable
 from functools import cached_property
-from typing import Any, Generic, TypeVar
+from typing import Any, Unpack
```
Tomas2D, line 7: `typing_extensions` again

python/beeai_framework/agents/base.py
```diff
@@ -1,28 +1,85 @@
...
+        """Execute an agent, specializing the runnable interface to accept a string as input
+        in addition to a list of messages.
+
+        Args:
+            input: The input to the agent
+            signal: The agent abort signal
+            context: A dictionary that can be used to pass additional context to the agent
+
+        Returns:
+            The agent output.
```
Tomas2D, line 80: I don't think this is needed since it is part of `AgentOptions` and visible to IDEs.

python/beeai_framework/agents/base.py
```diff
@@ -85,3 +138,25 @@ async def handler(context: RunContext) -> TOutput:
...
+#         """Convert list[AnyMessage] to str by taking the last element."""
+#         self = args[0] if args else None
+#         if not isinstance(self, BaseAgent):
+#             raise TypeError("The first argument of an Agent's run method must be an Agent instance.")
+#         input: Any = args[1] if args and len(args) > 1 else None
+#         text_input = input if isinstance(input, str) else (input[-1].text if input else "")
+#         new_args = (args[0], text_input) + args[2:]
+#         return func(*new_args, **kwargs)  # type: ignore[arg-type]  # We preserve the original arg structure
+
+#     return wrapper
```
Tomas2D: Can it be removed?

python/beeai_framework/agents/experimental/agent.py
```diff
@@ -98,22 +102,31 @@ def __init__(
...
+            expected_output: Pydantic model or instruction for steering the agent towards an expected output format.
+            total_max_retries: Maximum number of model retries.
+            max_retries_per_step: Maximum number of model retries per step.
+            max_iterations: Maximum number of iterations.
+            backstory: Additional piece of information or background for the agent.
+            signal: The agent abort signal
+            context: A dictionary that can be used to pass additional context to the agent
+
+        Returns:
+            The agent output.
```
Tomas2D, line 123: Duplicated as everything has already been defined in `AgentOptions`.

python/beeai_framework/agents/experimental/agent.py
```diff
@@ -98,22 +102,31 @@ def __init__(
...
+        """
+        text_input = input if isinstance(input, str) else (input[-1].text if input else "")
```
Tomas2D:
```
text_input = input[-1].text if input and isinstance(input, str) and isinstance(input[-1], UserMessage) else input
```

1. simplified
2. checks for user message only

python/beeai_framework/agents/experimental/agent.py
```diff
@@ -98,22 +102,31 @@ def __init__(
...
+        run_config = AgentExecutionConfig(
+            max_retries_per_step=kwargs.get("max_retries_per_step", 3),
+            total_max_retries=kwargs.get("total_max_retries", 3),
+            max_iterations=kwargs.get("max_iterations", 20),
```
Tomas2D: Idea: Maybe we could move `AgentExecutionConfig` from `types.py` to `base.py` and add a static method like `.from_options(...)` that would do this for us.

python/beeai_framework/agents/experimental/agent.py
```diff
@@ -264,20 +277,9 @@ async def handler(run_context: RunContext) -> RequirementAgentRunOutput[TOutput]
...
-            handler,
-            signal=signal,
-            run_params={
-                "prompt": prompt,
-                "context": context,
-                "expected_output": expected_output,
-                "execution": execution,
-            },
-        )
+        return await handler(RunContext.get())
```
Tomas2D, line 311: We should remove the handler wrapper in a follow-up PR, but I assume you did it this way to avoid too many changes at once and to keep the diff clear 👍🏻

python/beeai_framework/agents/experimental/rag/agent.py
```diff
@@ -1,19 +1,21 @@
 # Copyright 2025 © BeeAI a Series of LF Projects, LLC
 # SPDX-License-Identifier: Apache-2.0
 
+from typing import Unpack
```
Tomas2D, line 4: `typing_extensions`

python/beeai_framework/agents/experimental/rag/agent.py
```diff
@@ -49,23 +51,31 @@ def _create_emitter(self) -> Emitter:
...
+
+        Args:
+            input: The input to the agent (if list of messages, uses the last message as input)
+            total_max_retries: Maximum number of model retries.
+            signal: The agent abort signal
+            context: A dictionary that can be used to pass additional context to the agent
+
+        Returns:
+            The agent output.
+        """
```
Tomas2D, line 56: Unnecessary + I can see that the docstring is actually outdated.

python/beeai_framework/agents/react/agent.py
```diff
@@ -102,107 +105,104 @@ def meta(self) -> AgentMeta:
...
+            total_max_retries: Maximum number of model retries.
+            max_retries_per_step: Maximum number of model retries per step.
+            max_iterations: Maximum number of iterations.
+            signal: The agent abort signal
+            context: A dictionary that can be used to pass additional context to the agent
+
+        Returns:
+            The agent output.
+        """
+        text_input = input if isinstance(input, str) else (input[-1].text if input else "")
```
Tomas2D: you are ignoring all messages, but the last one

Tomas2D: all messages should be passed to the memory, but the last one (if from the user should handled this way)

python/beeai_framework/agents/tool_calling/__init__.py
```diff
@@ -3,11 +3,11 @@
 
 from beeai_framework.agents.tool_calling.agent import ToolCallingAgent
 from beeai_framework.agents.tool_calling.events import ToolCallingAgentStartEvent, ToolCallingAgentSuccessEvent
-from beeai_framework.agents.tool_calling.types import ToolCallingAgentRunOutput, ToolCallingAgentTemplateFactory
+from beeai_framework.agents.tool_calling.types import ToolCallingAgentOutput, ToolCallingAgentTemplateFactory
 
 __all__ = [
     "ToolCallingAgent",
-    "ToolCallingAgentRunOutput",
+    "ToolCallingAgentOutput",
```
Tomas2D, line 14: breaking change again

python/beeai_framework/adapters/acp/agents/__init__.py
```diff
@@ -3,11 +3,11 @@
...
 from beeai_framework.adapters.acp.agents.agent import ACPAgent
 from beeai_framework.adapters.acp.agents.events import ACPAgentErrorEvent, ACPAgentUpdateEvent
-from beeai_framework.adapters.acp.agents.types import ACPAgentRunOutput
+from beeai_framework.adapters.acp.agents.types import ACPAgentOutput
 
 __all__ = [
     "ACPAgent",
     "ACPAgentErrorEvent",
-    "ACPAgentRunOutput",
+    "ACPAgentOutput",
```
Tomas2D, line 12: breaking change

python/beeai_framework/agents/react/__init__.py
```diff
@@ -9,13 +9,13 @@
     ReActAgentSuccessEvent,
     ReActAgentUpdateEvent,
 )
-from beeai_framework.agents.react.types import ReActAgentRunOutput, ReActAgentTemplateFactory
+from beeai_framework.agents.react.types import ReActAgentOutput, ReActAgentTemplateFactory
 
 __all__ = [
     "ReActAgent",
     "ReActAgentErrorEvent",
+    "ReActAgentOutput",
```
Tomas2D, line 18: breaking change

python/examples/agents/custom_agent.py
```diff
@@ -68,23 +56,23 @@ class CustomSchema(BaseModel):
                 messages=[
                     SystemMessage("You are a helpful assistant. Always use JSON format for your responses."),
                     *(self.memory.messages if self.memory is not None else []),
-                    run_input.message,
+                    UserMessage(text_input),
```
Tomas2D: This can be tricky because you might be interpreting a non-user message as a user message

Tomas2D (changes requested):
We might consider using `Field(..., description="...")` instead of just docstring when using Pydantic models, as it might be useful when converting the model to JSON Schema.

python/beeai_framework/adapters/a2a/agents/agent.py
xjacka: This shouldn't be a problem. We just lost a little information.

python/beeai_framework/adapters/a2a/agents/agent.py
araujof: I checked and it's supported in standard `typing` from [3.11 onwards](https://docs.python.org/3/library/typing.html#typing.Unpack). This is why the linters are okay with it.
I also tested it on 3.11 on my machine, since it's the minimum we require in BeeAI.

python/beeai_framework/adapters/a2a/agents/__init__.py
araujof: I added type aliases with a deprecation comment to all agent output types. We can probably drop them in 0.2.

python/beeai_framework/adapters/a2a/agents/agent.py
araujof: Thanks. Fixed.

python/examples/agents/custom_agent.py
araujof: True. I changed it to preserve the message types in case the input is already a list of messages. 

I think that any external messages that the agent receives should probably be terminated by a `USER` role in the framework. For example, we don't differentiate between a bot and a user. A non-human is still a user of that agent.

python/beeai_framework/adapters/a2a/agents/agent.py
araujof: Fixed.

python/beeai_framework/agents/react/agent.py
araujof: I fixed that for all agents. For the ReAct agent, I updated the runner to handle that.

python/beeai_framework/adapters/a2a/agents/agent.py
```diff
@@ -131,8 +131,8 @@ async def handler(context: RunContext) -> A2AAgentOutput:
                         self._reference_task_ids.append(self._task_id)
 
                     # add input message to memory
-                    input_message: AnyMessage = self._convert_message_to_framework_message(input)
-                    await self.memory.add(input_message)
+                    input_messages = self._convert_messages_to_framework_messages(input)
+                    await self.memory.add_many(input_messages)
```
Tomas2D: I think this method is not needed. You could just do this instead.

```python
messages = [self._convert_message_to_framework_message(msg) for msg in cast_list(input)]
```

python/beeai_framework/adapters/a2a/agents/types.py
```diff
@@ -1,10 +1,15 @@
...
 from beeai_framework.agents import AgentOutput
 
 
 class A2AAgentOutput(AgentOutput):
     event: Any
+
+
+# Deprecated: Use 'A2AAgentOutput' instead.
+# This alias will be removed in version 0.2
+A2AAgentRunOutput: TypeAlias = A2AAgentOutput
```
Tomas2D: Would you happen to know if the user will see this warning?

python/beeai_framework/agents/experimental/agent.py
```diff
@@ -133,6 +135,8 @@ async def init_state() -> tuple[RequirementAgentRunState, UserMessage | None]:
                 memory=UnconstrainedMemory(), steps=[], iteration=0, answer=None, result=None
             )
             await state.memory.add_many(self.memory.messages)
+            if isinstance(input, list):
+                await state.memory.add_many(input[:-1])
```
Tomas2D: If the list has just one message, you duplicate it.

python/beeai_framework/agents/react/runners/default/runner.py
```diff
@@ -325,13 +325,15 @@ async def _init_memory(self, input: ReActAgentRunInput) -> BaseMemory:
...
+        created_at = datetime.datetime.now(tz=datetime.UTC)
+        if input.prompt:
+            last_msg = input.prompt if isinstance(input.prompt, str) else input.prompt[-1].text
+            content = self.templates.user.render(UserPromptTemplateInput(input=last_msg, created_at=created_at))
+            if isinstance(input.prompt, list):
+                await memory.add_many(input.prompt[:-1])
+            await memory.add(UserMessage(content=content, meta={"createdAt": created_at}))
+        else:
+            content = self.templates.user_empty.render(UserEmptyPromptTemplateInput())
             await memory.add(UserMessage(content=content, meta={"createdAt": created_at}))
```
Tomas2D, line 343: The rewrite does not seem equivalent to me

python/beeai_framework/agents/tool_calling/agent.py
```diff
@@ -93,7 +93,10 @@ async def run(self, input: str | list[AnyMessage], /, **kwargs: Unpack[AgentOpti
         Returns:
             The agent output.
         """
+        if not input:
+            raise ValueError("Invalid input. The input must be a non-empty string or list of messages.")
```
Tomas2D: Also, breaking change. Empty input is valid.

Tomas2D (changes requested):
I am not sure whether the handling of adapters is correct. There are so many extra validation checks that should pass. I don't see imported updates for `Unpack`.

python/beeai_framework/adapters/a2a/agents/types.py
araujof: No. In Python 3.13+, typing.deprecated can be applied to type aliases, which mypy can detected and it would also log the warning. However, since we support minimum of Python 3.11, the only way here would be for us to define a wrapper utility that uses getatrr and log the deprecation.

python/beeai_framework/agents/experimental/agent.py
araujof: Hey, I don't think so. I'm using `:`. See, for, example:
```Python
l = [1, 2]
print(l[:-1])
>>> [1]

l = [1]
print(l[:-1])
>>> []
```

I'm marking this as resolved, unless you see some other issue. Thanks!

python/beeai_framework/adapters/a2a/serve/agent_executor.py
```diff
@@ -78,11 +76,13 @@ async def execute(
 
         await updater.start_work()
         try:
-            response = await self._agent.run(signal=self._abort_controller.signal)
+            response = await self._agent.run(
+                [convert_a2a_to_framework_message(context.message)], signal=self._abort_controller.signal
```
xjacka: The `context.message` is already present in the history.

python/beeai_framework/adapters/acp/serve/server.py
```diff
@@ -133,7 +133,7 @@ async def run(
         agent.memory.reset()
         await agent.memory.add_many(acp_msgs_to_framework_msgs(input))
 
-        async for data, event in agent.run():
+        async for data, event in agent.run(acp_msgs_to_framework_msgs(input)):
```
xjacka: The message is already in the memory.

python/beeai_framework/adapters/a2a/agents/types.py
araujof: Marking this as resolved for now. We may want to think more broadly how we want to handle deprecations in the framework. Should we open an issue for that?

The `deprecated` package works great for classes and methods, but not for other symbols. Python 3.13 introduces the new typing.deprecated with supported for type alias, etc., but we can't rely on it yet. Microsoft also has a linter that has some support for type alias deprecation, but when looking at it, it's not exactly what we want. 

So, for now, I'm proposing we do something like this (plus comments): 

1. Add the following to a utility module. 

```Python
T = TypeVar("T")

def deprecated_alias(old_name: str, reason: str) -> Callable[[T], T]:
    """
    Decorator-style function to create deprecated aliases.

    Usage:
        A2AAgentRunOutput = deprecated_alias("A2AAgentRunOutput", "Use A2AAgentOutput instead")(A2AAgentOutput)
    """

    def decorator(new_type: T) -> T:

        logger.warning(
            f"{old_name} is deprecated" + (f". {reason}" if reason else ""),
            stacklevel=2
        )        
        return new_type
    return decorator

```

2. Use the decorator in our deprecated type aliases:

```Python
# Deprecated: Use 'ToolCallingAgentOutput' instead.
# This alias will be removed in version 0.2
ToolCallingAgentRunOutput = deprecated_alias("ToolCallingAgentRunOutput", "Use ToolCallingAgentOutput instead")(ToolCallingAgentOutput)
```

3. Log a warning if the alias is used:
```bash
python -c "from beeai_framework.agents.tool_calling.types import ToolCallingAgentRunOutput"
2025-08-22 15:23:55 | WARNING  | beeai_framework.agents.tool_calling.types:<module>:92 - ToolCallingAgentRunOutput is deprecated. Use ToolCallingAgentOutput instead
```

This should be compatible with Python 3.11+.

araujof:
> We might consider using `Field(..., description="...")` instead of just docstring when using Pydantic models, as it might be useful when converting the model to JSON Schema.

Sounds good. Let's tackle this once we normalize all data types across agents. We can probably do it in one swipe, focusing on the core models.

python/beeai_framework/adapters/a2a/agents/types.py
araujof: I think I figured out a way to warn users when they import deprecated type aliases. Given we support python 3.11+, we can do something like this (and the optional type checking alias is just to preserve typing when doing type checking, not required):
```Python
deprecated_type_alias(__name__, AgentOutput, "AgentRunOutput")
if TYPE_CHECKING: # This will only be seen by type checkers, not at runtime
    AgentRunOutput: TypeAlias = AgentOutput
```
Implementation of the wrappers: https://github.com/i-am-bee/beeai-framework/blob/8c3e87c036ece38afb79fa2ec87097d14d199cb3/python/beeai_framework/utils/warnings.py#L45

python/beeai_framework/agents/react/runners/default/runner.py
araujof: That's right, it's not equivalent. I updated the logic to reflect that we now require `input` (can't pass empty args to `run`), and that we also accept a list of messages, in addition to a string as input. 

We don't need to include an explicit case for empty memory, I think. The case when `prompt` is empty (`""` or `[]`) is included in the condition. I think the logic is correct (and adjusted based on our new contract for `run` and the input model for the runners), but please let me know if I missed anything.

python/beeai_framework/agents/tool_calling/agent.py
araujof: Do we want to still accept empty inputs in the agent implementations? If so, what should the treatment be? Maybe, something like this?

```Python
if not input and memory.is_empty():
    raise ValueError("Invalid input. The input must be a non-empty string or list of messages.")
```

python/beeai_framework/adapters/a2a/serve/agent_executor.py
araujof: I see, good catch. I that the adapter adds the messages to memory in bulk here:
https://github.com/i-am-bee/beeai-framework/blob/73fddff9056470cae0bab87f12dedeb2bd53ff08/python/beeai_framework/adapters/a2a/serve/agent_executor.py#L47-L58

However, the framework agents also update the memory internally when it receives a list of messages. I wonder if it would be okay to change the logic to just pass the list of messages to the agent in the `run` method (simpler, I think), and avoid updating the memory with the new messages during initialization? Wdyt?

python/beeai_framework/adapters/acp/serve/server.py
araujof: Same issue as in the a2a adapter. The agents now accept a list of messages as input, and they also update their internal memory with those messages, so maybe we need to change how we initialize the memory of the agents in these adapters.

python/beeai_framework/adapters/a2a/serve/agent_executor.py
araujof: Maybe something like this would work:
```
await init_agent_memory(...)
await self._agent.memory.add_many(
            [
                convert_a2a_to_framework_message(message)
                for message in (
                    (context.current_task.history or [])
                    if context.current_task
                    else []
                )
                if all(isinstance(part.root, a2a_types.TextPart) for part in message.parts)
            ]
        )
```

python/beeai_framework/adapters/a2a/agents/agent.py
araujof: Refactored with your suggestion. Thanks.

python/beeai_framework/agents/react/runners/default/runner.py
araujof: I refactored the code to bring back the memory empty check. I realize that some folks may still have legacy code that works in the old way... so for now, let's keep it this way. We need to update all our examples in the docs to use the input, and leverage the fact it takes a list of messages now. Thanks for pointing this out.

python/beeai_framework/agents/tool_calling/agent.py
araujof: I added the check above. That should be a good validation to have.

python/beeai_framework/agents/base.py
araujof: Ok, let's tackle this in another iteration.

python/beeai_framework/agents/base.py
araujof: This is in response to comments for other folks (Michael, Teryl, Anton, etc.)., that we should comment those kwargs well. So, let's err on the side of documenting them for this core interface, even though I understand the point.

python/beeai_framework/agents/experimental/agent.py
araujof: I was just trying to document the kwargs that this specific agent uses. Not all agents use all of the kwargs currently. 
Let's fix and normalize the docstrings in our next iteration. This one has become a bit too long already, and I feel like once we bring the agents together, a lot of this is going to become cleaner and simpler. :)

python/beeai_framework/agents/experimental/agent.py
araujof: got it, I think you meant:
```
text_input = input[-1].text if input and isinstance(input, list) and isinstance(input[-1], UserMessage) else input
```

araujof: updated it to check for UserMessage.

python/beeai_framework/agents/experimental/rag/agent.py
araujof: We talked about it and I checked, typing.Unpack in Python 3.11+ is the one to use. typing_extensions.Unpack is for Python <3.11.

python/beeai_framework/agents/experimental/rag/agent.py
araujof: Let's tackle the docstrings in our next PR.

python/beeai_framework/adapters/a2a/agents/agent.py
araujof: I reverted it back for now... type checking issue... he cast_list returns a list[T]... 
```
error: Argument 1 to "_convert_message_to_framework_message" of "A2AAgent" has incompatible type
"str | list[beeai_framework.backend.message.Message[Any]] | beeai_framework.backend.message.Message[Any] | a2a.types.Message"; expected
"str | beeai_framework.backend.message.Message[Any] | a2a.types.Message | Artifact"  [arg-type]
                        messages = [self._convert_message_to_framework_message(msg) for msg in cast_list(input)]
```

python/beeai_framework/agents/experimental/agent.py
araujof: I like that, and though of something very similar. Let's push this change to the next PR.

python/beeai_framework/adapters/acp/serve/server.py
araujof: I think I addressed this. Please check and re-open the conversation in case you disagree.

python/beeai_framework/adapters/a2a/serve/agent_executor.py
araujof: I think I addressed this. Please check and re-open the conversation in case you disagree.

araujof:
@xjacka @Tomas2D Thanks much for the thorough review and comments! I think I addressed all points. 

On the memory initialization for agents (especially in the agent adapters), I refactored things a bit to address the duplication issues you pointed out. I also tried to improve the input and memory management for the core agents to work in situations where users still pass an empty input using the memory as input (for backwards compatibility). Please take a look. 

I think it's important that we try to prioritize this PR. The longer it stays as a branch, the bigger is the risk for major conflicts with changes in main, as it touches a lot of code. We can address other improvements in separate PRs.

araujof:
> I am not sure whether the handling of adapters is correct. There are so many extra validation checks that should pass. I don't see imported updates for `Unpack`.

@Tomas2D My validation checks are all passing. Am I missing anything? typing.Unpack = typing_extensions.Unpack for Python 3.11+. Let me know if I missed anything. I do see all the type hints on my IDE working properly for Unpack.

python/beeai_framework/adapters/a2a/agents/__init__.py
```diff
@@ -1,13 +1,26 @@
...
 from beeai_framework.adapters.a2a.agents.agent import A2AAgent
 from beeai_framework.adapters.a2a.agents.events import A2AAgentErrorEvent, A2AAgentUpdateEvent
-from beeai_framework.adapters.a2a.agents.types import A2AAgentRunOutput
+from beeai_framework.adapters.a2a.agents.types import A2AAgentOutput
+
+
+# lazy import deprecated aliases to prevent unecessary warnings
+def __getattr__(name: str) -> Any:
+    if name == "A2AAgentRunOutput":
+        from beeai_framework.adapters.a2a.agents.types import A2AAgentRunOutput
```
Tomas2D: There should be a deprecation warning.

```
warnings.warn(
            "A2AAgentRunOutput is deprecated, use A2AAgentOutput instead.",
            DeprecationWarning,
            stacklevel=2,
        )
```

python/beeai_framework/workflows/agent/agent.py
```diff
@@ -34,16 +33,16 @@
 
 
 class AgentWorkflowInput(BaseModel):
-    prompt: str | None = None
-    context: str | None = None
+    input: str
+    backstory: str | None = None
```
Tomas2D: I don't think the breaking change is necessary in something we will soon remove.

python/examples/run.py
Tomas2D, line 1: This file is not needed. Please remove.

One can run the example script directly, either from the command line or from the IDE, in which `examples` can be marked as Sources Root. In addition, this example contains an additional dependency `runpy`.

If you'd like to keep it, just extend the `task.toml` file with an additional script (it would just inject the `examples` directory path into the env).

python/beeai_framework/adapters/watsonx_orchestrate/serve/_factories/_requirement_agent.py
```diff
@@ -24,11 +24,20 @@ class WatsonxOrchestrateServerRequirementAgent(WatsonxOrchestrateServerAgent[Req
...
-    async def _run(self) -> AssistantMessage:
-        response = await self._agent.run(prompt=None)
-        return response.answer
+    async def _run(self, input: list[AnyMessage]) -> AssistantMessage:
+        if input or not self._agent.memory.is_empty():
+            input = self._agent.memory.messages[-1:]
+            response = await self._agent.run(input)
+            return response.message
+        else:
+            raise ValueError("Agent invoked with empty memory.")
```
Tomas2D: Why did you change the signature of the function? It should be kept as is. If you read the flow carefully, you will see the memory is set before the `run` is called. Please revert your changes, as this completely breaks the agent.

python/beeai_framework/adapters/watsonx_orchestrate/serve/_factories/_requirement_agent.py
```diff
@@ -24,11 +24,20 @@ class WatsonxOrchestrateServerRequirementAgent(WatsonxOrchestrateServerAgent[Req
...
+
+    async def _stream(self, input: list[AnyMessage], emit: WatsonxOrchestrateServerAgentEmitFn) -> None:
+        if not input and self._agent.memory.is_empty():
+            raise ValueError("Agent invoked with empty memory.")
+
+        input = self._agent.memory.messages[-1:]
```
Tomas2D: same

python/beeai_framework/adapters/watsonx_orchestrate/serve/_factories/_requirement_agent.py
Tomas2D: https://github.com/i-am-bee/beeai-framework/blob/df8a4c0b49d6a48aca568097411390b017faaf83/python/beeai_framework/adapters/watsonx_orchestrate/serve/api.py#L78-L90

python/beeai_framework/adapters/watsonx_orchestrate/serve/_factories/_tool_calling_agent.py
```diff
@@ -21,11 +21,20 @@ class WatsonxOrchestrateServerToolCallingAgent(WatsonxOrchestrateServerAgent[Too
...
+            response = await self._agent.run(input)
+            return response.message
+        else:
+            raise ValueError("Agent invoked with empty memory.")
+
+    async def _stream(self, input: list[AnyMessage], emit: WatsonxOrchestrateServerAgentEmitFn) -> None:
+        if not input and self._agent.memory.is_empty():
+            raise ValueError("Agent invoked with empty memory.")
+
+        input = self._agent.memory.messages[-1:]
```
Tomas2D: same

python/beeai_framework/agents/base.py
```diff
@@ -1,28 +1,85 @@
...
-        super().__init__()
+    expected_output: str | type[BaseModel]
+    """
+    Instruction for steering the agent towards an expected output format.
+    This can be a Pydantic model for structured output decoding and validation.
+    """
+
+    total_max_retries: int
+    """
+    Maximum number of model retries.
```
Tomas2D: This is not bound to LLM (model), but it includes everything that can be retried. Use a more generic word, like `step`, so that one can see the connection to `max_retries_per_step`.

python/beeai_framework/agents/base.py
```diff
@@ -1,28 +1,85 @@
...
+    """
+
+    max_retries_per_step: int
+    """
+    Maximum number of model retries per step.
+    """
+
+    max_iterations: int
+    """
+    Maximum number of iterations.
```
Tomas2D, line 43: Either explain what iteration is to the user or remove the comment, as it adds no value.

python/beeai_framework/agents/base.py
```diff
@@ -1,28 +1,85 @@
...
+    """
+    Additional piece of information or background for the agent.
+    """
+
+
+class AgentOutput(RunnableOutput):
+    """Agent output."""
+
+    output_structured: Any | BaseModel = None
+    """The formatted output returned by the agent."""
```
Tomas2D, line 56: Show an example of what that is. Some users might know the concept.

python/beeai_framework/agents/base.py
Tomas2D: Sorry, could you rephrase the point? I understand commenting on the intent of the function, but not the kwargs in this specific case.

python/beeai_framework/agents/base.py
```diff
@@ -57,17 +110,24 @@ def meta(self) -> AgentMeta:
             tools=[],
         )
 
+    def run_config(self, **kwargs: Unpack[AgentOptions]) -> AgentExecutionConfig:
```
Tomas2D: What is it? The name is very confusing; it is called run, but it does not run anything. We don't want to have dozens of methods (especially public). 

If you wanted to address my prior comment, I was talking about adding a factory function (in utils) or factory static method to the `AgentExecutionConfig`.

python/beeai_framework/agents/experimental/__init__.py
```diff
@@ -1,7 +1,22 @@
...
 from beeai_framework.agents.experimental.agent import RequirementAgent
-from beeai_framework.agents.experimental.types import RequirementAgentRunOutput, RequirementAgentRunState
+from beeai_framework.agents.experimental.types import (
+    RequirementAgentOutput,
+    RequirementAgentRunState,
+)
+
+
+# lazy import deprecated aliases to prevent unecessary warnings
+def __getattr__(name: str) -> Any:
```
Tomas2D: Consider removing all deprecating fallbacks and including a migration guide in the PR instead so that we can announce it as part of the release.

python/beeai_framework/agents/react/runners/default/runner.py
```diff
@@ -325,13 +325,15 @@ async def _init_memory(self, input: ReActAgentRunInput) -> BaseMemory:
...
-                else self.templates.user_empty.render(UserEmptyPromptTemplateInput())
-            )
+        created_at = datetime.datetime.now(tz=datetime.UTC)
+        if input.prompt:
+            last_msg = input.prompt if isinstance(input.prompt, str) else input.prompt[-1].text
+            content = self.templates.user.render(UserPromptTemplateInput(input=last_msg, created_at=created_at))
+            if isinstance(input.prompt, list):
+                await memory.add_many(input.prompt[:-1])
+            await memory.add(UserMessage(content=content, meta={"createdAt": created_at}))
+        elif memory.is_empty():
```
Tomas2D: 1. It seems that `memory.is_empty()` is always false, because it contains the System message. I would change it to `if len(memory.messages) < 2` (`elif` not needed)
2. `last_msg = input.prompt if isinstance(input.prompt, str) else input.prompt[-1].text` should only apply if the input is `str`, if it is a `list`, just append all messages and don't change anything

python/beeai_framework/agents/react/types.py
```diff
@@ -18,17 +19,19 @@
...
 from beeai_framework.memory.base_memory import BaseMemory
 from beeai_framework.template import PromptTemplate
 from beeai_framework.tools.tool import AnyTool
 from beeai_framework.utils.strings import to_json
+from beeai_framework.utils.warnings import deprecated_type_alias
 
 
 class ReActAgentRunInput(BaseModel):
-    prompt: str | None = None
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
Tomas2D: `arbitrary_types_allowed` is not needed, use `prompt: str | list[InstanceOf[AnyMessage]]` instead

python/beeai_framework/adapters/a2a/agents/__init__.py
Tomas2D: or as proposed later on ... let's maybe remove all of this and just introduce a friendly "migration guide" as there are more changes

araujof:
@xjacka can you help us run integration tests with the A2A/ACP/Adapters? Thanks much!

python/beeai_framework/agents/react/runners/default/runner.py
araujof: Question: do we need to check if the last message in the input is a UserMessage, and that case, format it with the template, like here:
```
content = self.templates.user.render(UserPromptTemplateInput(input=last_msg, created_at=created_at))
```

python/beeai_framework/agents/experimental/__init__.py
araujof: Should we keep the export of the deprecated aliases in the init modules? The idea of adding the fallbacks is so that we could emit warnings on them in case they get imported by consumers/users of the framework.

python/beeai_framework/agents/base.py
araujof: Right, agree. Let me change it.

python/beeai_framework/adapters/watsonx_orchestrate/serve/_factories/_requirement_agent.py
araujof: I refactored the memory initialization to just memory creation (without adding the messages). I tried to explain this in my comment to @xjacka in this PR. Since the agent now accepts a list of messages, I think it's probably best to let the agent handle how it updates its memory (memory management). It also reduces the chances of having duplicate messages being added to their memory.

python/examples/run.py
araujof: I added it there for convenience. Many of the examples don't load the .env file. I can remove it, just thought it was nice to have when you wanted a quick way to interact and debug a specific example.

python/beeai_framework/workflows/agent/agent.py
araujof: Undone.

python/beeai_framework/agents/experimental/agent.py
```diff
@@ -98,42 +102,61 @@ def __init__(
...
-        context: str | None = None,
-        expected_output: str | type[TOutput] | None = None,
-        execution: AgentExecutionConfig | None = None,
-        signal: AbortSignal | None = None,
-    ) -> Run[RequirementAgentRunOutput[TOutput]]:
-        run_config = execution or AgentExecutionConfig(
-            max_retries_per_step=3,
-            total_max_retries=3,
-            max_iterations=20,
+    @runnable_entry
```
antonpibm, line 108: If the base agent inherits Runnable, why do you need to annotate all run methods as `runnable_entry`? shouldn't it be the default behavior? this comment applies to all agent implementations

python/beeai_framework/agents/experimental/agent.py
```diff
@@ -98,42 +102,61 @@ def __init__(
...
+                    context=kwargs.get("backstory"),
                     expected_output=expected_output if isinstance(expected_output, str) else None,  # TODO: validate
                 )
                 user_message = UserMessage(self._templates.task.render(task_input))
                 await state.memory.add(user_message)
 
             return state, user_message
 
-        async def handler(run_context: RunContext) -> RequirementAgentRunOutput[TOutput]:
+        async def handler(run_context: RunContext) -> RequirementAgentOutput:
```
antonpibm, line 168: Now that the run method is async and returns the same type as the handler, is `async def handler` method still needed?

araujof:
@xjacka @jenna-winkler I just spoke to @Tomas2D - we will wait for #1004 to be merged, rebased this PR (resolving conflicts), and then should be good to go for a final review and approval.

python/beeai_framework/agents/experimental/agent.py
araujof: No, it's not. We will drop that in a separate PR. Just trying to minimize the amount of potential conflicts with the main branch at this point :)

python/beeai_framework/agents/experimental/agent.py
araujof: This would work, if we refactor all agents to update their signatures. Could probably generate some bigger breaking changes for folks who have subclassed BaseAgent already in their code, but not sure this is a big issue here. Maybe we tackle this one in a future PR? This one is already too big...

araujof:
@Tomas2D I merged from `main` and resolved the conflicts introduced in #1004. Can you please review the changes to A2A and platform adapters?

@xjacka could you please help me test these adapters to make sure I didn't introduce any regressions? Thanks much!

python/examples/serve/acp_with_custom_agent.py
```diff
@@ -38,17 +35,17 @@ def _create_emitter(self) -> Emitter:
...
             assert self.memory is not None
-            if input:
-                await self.memory.add_many(input)
-            return EchoAgentRunOutput(message=AssistantMessage(self.memory.messages[-1].text))
+            if text_input:
+                await self.memory.add(UserMessage(content=text_input))
+            return AgentOutput(output=[AssistantMessage(text_input)])
 
-        return self._to_run(handler, signal=None)
+        return await handler(RunContext.get())
```
Tomas2D: The wrapping here is no longer needed.

araujof: Right. How about I do a full pass over these once this PR is merged? I can do it right after. 
Point is that we probably want to reduce conflict resolution issues at this point.

Tomas2D:
I reverted `AgentWorkflowInput`, merged main, and improved the RunnableOutput definition.

Tomas2D (approved):

Issue #1025, opened 2025-08-22
Workflow V2 Implementation


Tomas2D:
Ref: #1005 (once done)

sandijean90:
Some things are being worked through by @michael-desmond :
- There are Typing limitations that haven't been fixed in base python. Figuring out a workaround that doesn't make the the piping system brittle. 
- Deciding whether to use the runnable rather than make the workflow run a class, but then need to figure out how to pass the state. Also figuring out how things come back together to a single step after two having been ran in parallel before it.

PR #1124, opened 2025-09-22
feat: workflows v2

michael-desmond:

Addresses #1025 

### Description

Lightweight agent orchestration using decorators.

#### General

#### Code quality checks

#### Testing
- [ ] Unit tests pass: `mise test:unit`

#### Documentation
- [ ] Embedme embeds code examples in docs. To update after edits, run: Python `mise docs:fix`


python/beeai_framework/workflows/v3/step.py
```diff
@@ -0,0 +1,198 @@
...
+    async def execute(self, *inputs: Any) -> Any:
+        pass
+
+    @property
+    def result(self) -> Any:
+        return None
+
+    async def requeue(self, *inputs: Any) -> bool:
+        if self.loop_condition:
+            return bool(await run_callable(self.loop_condition.fn, inputs))
```
Tomas2D: Shouldn't `inputs` be expanded?

```python
result = await run_callable(self.loop_condition.fn, *inputs)
return bool(result)
```

python/beeai_framework/workflows/v3/workflow.py
```diff
@@ -0,0 +1,196 @@
...
+
+            if tasks:
+                # done, _ = await asyncio.wait(tasks)
+                done, _ = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
+
+        output: RunnableOutput = (
+            completed_steps[-1].result
+            if completed_steps and isinstance(completed_steps[-1].result, RunnableOutput)
+            else RunnableOutput(output=[])
+        )
```
Tomas2D, line 150: In parallel/branching graphs, this is nondeterministic. Workflows with multiple terminal leaves may not return the intended value. Is this handled?

python/beeai_framework/workflows/v3/workflow.py
```diff
@@ -0,0 +1,196 @@
...
+                if enqueue_ds and ds_edge.condition is not None:
+                    if ds_edge.condition.fn:
+                        enqueue_ds = bool(await run_callable(ds_edge.condition.fn, *results) == ds_edge.condition.key)
+                    else:
+                        enqueue_ds = bool(step.result == ds_edge.condition.key)
+
+                if enqueue_ds:
+                    ds.has_executed = False
+                    await queue.put(ds)
+
```
Tomas2D, line 132: If a downstream step has multiple upstream edges marked as or, each upstream completion may enqueue the downstream, causing multiple concurrent executions and potential state races.

python/beeai_framework/workflows/v3/workflow.py
```diff
@@ -0,0 +1,196 @@
...
+            while not queue.empty():
+                step = await queue.get()
+                task = asyncio.create_task(execute_step(step))
+                tasks.add(task)
+                task.add_done_callback(tasks.discard)
+                queue.task_done()
+
+            if tasks:
+                # done, _ = await asyncio.wait(tasks)
+                done, _ = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
```
Tomas2D, line 144: I think that other tasks should be awaited or cancelled.

python/beeai_framework/workflows/v3/workflow.py
```diff
@@ -0,0 +1,196 @@
...
+    """
+    Descriptor that turns an async method into a WorkflowStep.
+    Users can refer to decorated methods and treat as WorkflowStep for composition.
+    """
+
+    def __init__(self, func: AsyncStepFunction, is_start: bool = False, is_end: bool = False) -> None:
+        self.func = func
+        self.name = func.__name__
+        self.is_start = is_start
+        self.is_end = is_end
```
Tomas2D: Unused. Can be removed.

python/beeai_framework/workflows/v3/workflow.py
```diff
@@ -0,0 +1,196 @@
...
+            return self  # accessed on class, not instance
+
+        cache_name = f"__workflow_step_cache_{self.name}"
+        if not hasattr(instance, cache_name):
+            bound_func = self.func.__get__(instance, owner)
+            setattr(instance, cache_name, self.step_factory(func=bound_func))
+        return getattr(instance, cache_name)
+
+    def step_factory(self, func: AsyncStepFunction) -> WorkflowStep:
+        return FuncWorkflowStep(func=func)
```
Tomas2D: Does not seem to be needed.

python/beeai_framework/workflows/v3/workflow.py
```diff
@@ -0,0 +1,196 @@
...
+    WorkflowBuilder,
+    WorkflowStep,
+)
+from beeai_framework.workflows.v3.types import AsyncStepFunction
+from beeai_framework.workflows.v3.util import run_callable
+
+T = TypeVar("T", bound="Workflow")
+
+
+class step:  # noqa: N801
```
Tomas2D, line 29: This could be a function instead of a class.

python/beeai_framework/workflows/v3/workflow.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def __set_name__(self, owner: T, name: str) -> None:
+        self.name = name
+
+    @overload
+    def __get__(self, instance: None, owner: type[T]) -> "step": ...
+    @overload
+    def __get__(self, instance: T, owner: type[T]) -> WorkflowStep: ...
+
+    def __get__(self, instance: T | None, owner: type[T]) -> Any:
```
Tomas2D, line 47: These overloads might not be needed as we are no longer using `is_start` etc.

python/beeai_framework/workflows/v3/workflow.py
michael-desmond: Using the `.after` operator it is possible to build a workflow with multiple terminal steps, in which case you get an empty RunnableOutput from the outer workflow instance. Its up to the programmer of the workflow to properly join to a terminal step that returns RunnableOutput.

python/beeai_framework/workflows/v3/workflow.py
michael-desmond: Yes but the Workflow builder is designed to prevent this situation so that the workflow is predictable. I.e. parallel steps have an implicit downstream join step with `and` edges. And you cant call `.then` after `.branch` because you would be joining multiple upstream into a single downstream. This is the tricky part of the entire system.

python/beeai_framework/workflows/v3/workflow.py
michael-desmond: If you await all of the tasks here then the system will synchronize to the current level in the queue and it wont be able to queue downstream steps that can safely be executed concurrently.

python/beeai_framework/workflows/v3/workflow.py
michael-desmond: These overloads are necessary so that mypy wont get upset about the class being step but the return value of the __get__ being a WorkflowStep.

Tomas2D:
Not Planned

Issue #1145, opened 2025-09-26
Gemini: `MALFORMED_FUNCTION_CALL` clashes with tool response assertion


nforro:
**Describe the bug**
A while ago I reported an issue with Gemini models sometimes returning `MALFORMED_FUNCTION_CALL` (https://github.com/i-am-bee/beeai-framework/issues/959) that has been promptly fixed (thanks a lot). However, later a check for ignored tool choice requirement (https://github.com/i-am-bee/beeai-framework/pull/977) has been introduced and unfortunately it means that when Gemini responds with said error, despite supporting `tool_choice={"required"}`, the check is activated (because an effectively empty `ChatModelOutput` instance is generated, without any tool calls):

https://github.com/i-am-bee/beeai-framework/blob/edac7e338593bbf6a967709d512bffd3eb66643e/python/beeai_framework/backend/chat.py#L436-L441

Without the check in place the model eventually calls the tool properly and responds with a proper message.

I'm not sure if it is even possible to fix this properly. If not, would it be possible make the tool response assertion configurable?

**Expected behavior**
No error.

**Set-up:**

- Bee version: v0.1.43
- Model provider: Gemini

Tomas2D:
Hello @nforro. Can you please update to the latest BeeAI Framework version (0.1.50) and verify that the error is still present?

nforro:
> Hello [@nforro](https://github.com/nforro). Can you please update to the latest BeeAI Framework version (0.1.50) and verify that the error is still present?

Yes, it is still present.

```
DEBUG:LiteLLM:RAW RESPONSE:
{
  "candidates": [
    {
      "finishReason": "MALFORMED_FUNCTION_CALL",
      "index": 0
    }
  ],
  "usageMetadata": {
    "promptTokenCount": 103310,
    "totalTokenCount": 103310,
    "promptTokensDetails": [
      {
        "modality": "TEXT",
        "tokenCount": 103310
      }
    ]
  },
  "modelVersion": "gemini-2.5-pro",
  "responseId": "gNHWaJLTIe6ynsEPwIPY8Qo"
}

...

<-- 💬 GeminiChatModel[finish]: error has occurred
<-- 💬 GeminiChatModel[finish]: ChatModelError(beeai_framework.backend.errors): The model was required to produce a tool call, but no tool calls were generated.

[...]

PR #1203, opened 2025-10-18
Fix Gemini tool choice assertion when MALFORMED_FUNCTION_CALL occurs

mohin-io:
## Summary
- Fixes i-am-bee/beeai-framework#1145 by making the tool response assertion configurable
- default Gemini adapter to disable the assertion when Gemini returns MALFORMED_FUNCTION_CALL
- copy enforcement flag when cloning chat models and cover the behavior with regression tests

## Testing
- pytest -o addopts= python/tests/backend/test_chatmodel.py::test_chat_model_enforces_tool_choice_requirement *(blocked: missing 'litellm' dependency in runner)*
- pytest -o addopts= python/tests/backend/test_chatmodel.py::test_chat_model_tool_choice_assertion_can_be_disabled *(blocked: missing 'litellm' dependency in runner)*

### DIFF
--- a/python/beeai_framework/adapters/gemini/backend/chat.py
+++ b/python/beeai_framework/adapters/gemini/backend/chat.py
@@ -26,6 +26,8 @@ class GeminiChatModel(LiteLLMChatModel):
         api_key: str | None = None,
         **kwargs: Unpack[ChatModelKwargs],
     ) -> None:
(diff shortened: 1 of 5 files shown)


Tomas2D:
This functionality just suppresses the error. The agent will fail anyway, as it expects a valid tool call. Or do I miss something?

Tomas2D:
Suppressed by #1253.

PR #1205, opened 2025-10-18
Add Python serialization examples

### LINKED ISSUE #1200: Replace “Example coming soon” placeholders in serialization docs
**Problem:** 
docs/modules/serialization.mdx contains multiple “Example coming soon” placeholders (e.g., lines 22, 76, 114, 157, 206), leaving readers without runnable guidance on persistence APIs.

**What to do:**

- Produce Python and TypeScript snippets that mirror each described scenario (basic save/load, versioning, partial restores, encryption, etc.).
- Embed the examples in the doc using the existing <CodeGroup> pattern and link to source files under python/examples and typescript/examples.
- Validate instructions so users can copy, run, and see expected console output.

**Acceptance:** 

- Every placeholder is replaced with working code and brief narrative that clarifies prerequisites and expected results.

Sakeeb91:
## Summary
- add a lightweight `Serializer`/`Serializable` implementation for the Python SDK so the docs can point at runnable snippets
- replace the "Example coming soon" placeholders in `docs/modules/serialization.mdx` with concrete Python and TypeScript examples that follow existing codegroup conventions
- add focused pytest coverage for the new serializer behavior and surface the Python examples under `python/examples/serialization`

## Design Decisions
- kept the serializer API intentionally minimal (registry-based with JSON payloads) to mirror the TypeScript shape without pulling in heavy dependencies
- registered core backend message types lazily so docs examples (and deserialization) work without forcing global imports
- stored Python examples alongside existing samples to stay consistent with the docs embedding workflow

## Testing
- python/.venv/bin/python -m pytest python/tests/serialization/test_serializer.py
- python/.venv/bin/python -m pytest -m unit *(fails: missing optional deps `deepeval`, `outlines`)*

Closes #1200.

### DIFF
--- a/docs/modules/serialization.mdx
+++ b/docs/modules/serialization.mdx
@@ -21,6 +21,9 @@ The Python example prints the serialized payload, the restored ISO timestamp, an
 
 {/* <!-- embedme python/examples/serialization/base.py --> */}
 ```py Python [expandable]
+# Copyright 2025 © BeeAI a Series of LF Projects, LLC
+# SPDX-License-Identifier: Apache-2.0
+
 from datetime import UTC, datetime
 
 from beeai_framework.serialization import Serializer
@@ -38,6 +41,7 @@ def main() -> None:
 
 if __name__ == "__main__":
     main()
+
 ```
 
 {/* <!-- embedme typescript/examples/serialization/base.ts --> */}
@@ -93,6 +97,9 @@ Running the Python snippet restores the memory, appends an assistant reply, and 
 
 {/* <!-- embedme python/examples/serialization/memory.py --> */}
 ```py Python [expandable]
+# Copyright 2025 © BeeAI a Series of LF Projects, LLC
+# SPDX-License-Identifier: Apache-2.0
+
 import asyncio
 
 from beeai_framework.backend import AssistantMessage, UserMessage
@@ -114,6 +121,7 @@ async def main() -> None:
 
 if __name__ == "__main__":
     asyncio.run(main())
+
 ```
 
 {/* <!-- embedme typescript/examples/serialization/memory.ts --> */}
@@ -153,6 +161,9 @@ This example registers a lightweight data class and prints the reconstructed tok
 
 {/* <!-- embedme python/examples/serialization/custom_external.py --> */}
 ```py Python [expandable]
+# Copyright 2025 © BeeAI a Series of LF Projects, LLC
+# SPDX-License-Identifier: Apache-2.0
+
 from dataclasses import dataclass
 from datetime import UTC, datetime
 
@@ -189,6 +200,7 @@ def main() -> None:
 
 if __name__ == "__main__":
     main()
+
 ```
 
 {/* <!-- embedme typescript/examples/serialization/customExternal.ts --> */}
@@ -233,6 +245,9 @@ The Python variant increments a counter, serializes it, and prints both the live
 
 {/* <!-- embedme python/examples/serialization/custom_internal.py --> */}
 ```py Python [expandable]
+# Copyright 2025 © BeeAI a Series of LF Projects, LLC
+# SPDX-License-Identifier: Apache-2.0
+
 from beeai_framework.serialization import Serializable
 
 
@@ -263,6 +278,7 @@ def main() -> None:
 
 if __name__ == "__main__":
     main()
+
 ```
 
 {/* <!-- embedme typescript/examples/serialization/customInternal.ts --> */}
@@ -307,25 +323,29 @@ Failure to register a class that the `Serializer` does not know will result in t
 
 ## Context matters
 
-Deserialize with the `extraClasses` equivalent to make sure message factories are registered; the Python snippet prints every restored user utterance.
+Deserialize with the `extraClasses` equivalent to make sure message factories are registered; the Python snippet imports both `UserMessage` and `AssistantMessage` so the serializer can hydrate their content and then prints every restored user utterance.
 
 <CodeGroup>
 
 {/* <!-- embedme python/examples/serialization/context.py --> */}
 ```py Python [expandable]
-from beeai_framework.backend import UserMessage
+# Copyright 2025 © BeeAI a Series of LF Projects, LLC
+# SPDX-License-Identifier: Apache-2.0
+
+from beeai_framework.backend import AssistantMessage, UserMessage
 from beeai_framework.memory import UnconstrainedMemory
 
-SERIALIZED_MEMORY = ""
+SERIALIZED_MEMORY = """{"__version":"0.0.0","__root":{"__serializer":true,"__class":"UnconstrainedMemory","__value":{"messages":[{"__serializer":true,"__class":"UserMessage","__value":{"id":null,"meta":{"createdAt":{"__serializer":true,"__class":"datetime","__value":"2025-10-18T19:38:37.859543+00:00"}},"role":"user","content":[{"type":"text","text":"Hello!"}]}},{"__serializer":true,"__class":"AssistantMessage","__value":{"id":null,"meta":{"createdAt":{"__serializer":true,"__class":"datetime","__value":"2025-10-18T19:38:37.859665+00:00"}},"role":"assistant","content":[{"type":"text","text":"Hello, how can I help you?"}]}}]}}}"""
 
 
 def main() -> None:
-    memory = UnconstrainedMemory.from_serialized(SERIALIZED_MEMORY, extra_classes=[UserMessage])
+    memory = UnconstrainedMemory.from_serialized(SERIALIZED_MEMORY, extra_classes=[UserMessage, AssistantMessage])
     print([message.text for message in memory.messages])
 
 
 if __name__ == "__main__":
     main()
+
 ```
 
 {/* <!-- embedme typescript/examples/serialization/context.ts --> */}
--- a/docs/modules/serialization.mdx
+++ b/docs/modules/serialization.mdx
@@ -15,11 +15,29 @@ BeeAI framework provides robust serialization capabilities through its built-in 
 - 📦 Snapshots: Create point-in-time captures of component state
 - 🔧 Reconstruction: Rebuild objects from their serialized representation
 
+The Python example prints the serialized payload, the restored ISO timestamp, and a boolean confirming the round trip.
+
 <CodeGroup>
 
-{/* <!-- comingsoon python/examples/serialization/base.py --> */}
+{/* <!-- embedme python/examples/serialization/base.py --> */}
 ```py Python [expandable]
-Example coming soon
+from datetime import UTC, datetime
+
+from beeai_framework.serialization import Serializer
+
+
+def main() -> None:
+    original = datetime(2024, 1, 1, tzinfo=UTC)
+    serialized = Serializer.serialize(original)
+    restored = Serializer.deserialize(serialized, expected_type=datetime)
+
+    print(serialized)
+    print(restored.isoformat())
+    print(restored == original)
+
+
+if __name__ == "__main__":
+    main()
 ```
 
 {/* <!-- embedme typescript/examples/serialization/base.ts --> */}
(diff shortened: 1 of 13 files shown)


Sakeeb91:
Thanks for the thorough review! I pushed a follow-up that:

- adds reference tracking/`__ref` support to the Python serializer so shared instances and cycles survive a round trip (with new identity + cycle tests)
- tightens the backend message deserializer to validate the `'type'` field explicitly
- switches `Serializable.from_snapshot` to use `__new__` instead of assuming a no-arg `__init__`
- updates the context example to ship a real serialized payload and list both `UserMessage` and `AssistantMessage` in `extra_classes`

Let me know if you spot anything else!

sandijean90:
@Sakeeb91 thanks for your work here! We're excited to have you on board as a contributor. I've assigned to @Tomas2D to review/comment before merging.

docs/modules/serialization.mdx
```diff
@@ -199,11 +307,25 @@ Failure to register a class that the `Serializer` does not know will result in t
...
 <CodeGroup>
 
-{/* <!-- comingsoon python/examples/serialization/context.py --> */}
+{/* <!-- embedme python/examples/serialization/context.py --> */}
 ```py Python [expandable]
-Example coming soon
+from beeai_framework.backend import UserMessage
+
+SERIALIZED_MEMORY = ""
```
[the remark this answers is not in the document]
Tomas2D: Is this a valid comment? @Sakeeb91

python/beeai_framework/backend/message.py
```diff
@@ -374,3 +374,85 @@ def dedupe_tool_calls(msg: AssistantMessage) -> None:
 
     for idx in sorted(excluded_indexes, reverse=True):
         msg.content.pop(idx)
+
+
+def _register_message_class(
+    message_cls: type[Message[Any]],
+    allowed_content: Mapping[str, type[BaseModel]],
+) -> None:
+    def _registrar() -> None:
```
Tomas2D: typo I guess

python/beeai_framework/backend/message.py
```diff
@@ -374,3 +374,80 @@ def dedupe_tool_calls(msg: AssistantMessage) -> None:
...
+                "content": [fragment.model_dump() for fragment in message.content],
+            }
+
+        def _from_plain(payload: Mapping[str, Any]) -> Message[Any]:
+            content_payload = payload.get("content", []) or []
+            content_models: list[Any] = []
+
+            for item in content_payload:
+                if not isinstance(item, Mapping):
+                    raise SerializerError("Serialized message content must be a mapping.")
```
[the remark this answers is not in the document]
Tomas2D: Is this a valid comment? @Sakeeb91

python/beeai_framework/memory/token_memory.py
```diff
@@ -117,8 +118,9 @@ def reset(self) -> None:
         self._tokens_by_message.clear()
 
     async def clone(self) -> "TokenMemory":
+        llm_clone = self.llm.clone() if hasattr(self.llm, "clone") else self.llm
```
Tomas2D: Use `isistance(self.llm, Cloneable)` instead.

Tomas2D:
Thank you for your hard work! I am thinking if there is a more Pythonic way of doing that (pickling). What do you think?

Sakeeb91:
@Tomas2D thanks for the thoughtful suggestion! I explored pickle early on, but stuck with this JSON serializer for a couple reasons: it keeps deserialization side-effect free (pickle executes arbitrary code), the snapshots need to round-trip through our JavaScript SDK, and JSON is a lot easier to diff/debug in docs and tooling. With the reference tracking and tighter registry in this patch we close the identity gaps the review pointed out while staying interoperable.

python/beeai_framework/memory/token_memory.py
Sakeeb91: done!

Tomas2D:
I see your point. Thank you for the research. Before we merge this, can you verify that some basic examples work in both TS and Python so that one can use dump from TS and use it in Python, and vice versa?

Sakeeb91:
Verified the TS ↔︎ Python round-trip with runnable snippets:

 - **TS → Py:** `Serializer.serialize(UnconstrainedMemory)` snapshot restores via `UnconstrainedMemory.from_serialized(...)`, yielding the original `UserMessage`/`AssistantMessage` texts.
 - **Py → TS:** Python `memory.serialize()` payload hydrates in TypeScript by deserializing the memory and resolving each serializer node (same message types/text).

 Commands + outputs are in `reviewer_verification.md` (above). I also reran `poetry run pytest tests/serialization/test_serializer.py`. Let me know if you’d like the raw dumps or any additional coverage.

python/beeai_framework/serialization/serializer.py
```diff
@@ -0,0 +1,313 @@
...
+        instance.load_snapshot(snapshot)
+        return instance
+
+    def to_snapshot(self) -> T:
+        return self.create_snapshot()
+
+    def create_snapshot(self) -> T:
+        raise NotImplementedError
+
+    def load_snapshot(self, snapshot: T) -> None:
```
Tomas2D: I think that this should be:

1. abstract method
2. Awaitable

python/beeai_framework/serialization/serializer.py
```diff
@@ -0,0 +1,313 @@
...
+        assert isinstance(value, cls)
+        return value
+
+    @classmethod
+    def from_snapshot(cls: type[SerializableT], snapshot: T) -> SerializableT:
+        instance = object.__new__(cls)
+        instance.load_snapshot(snapshot)
+        return instance
+
+    def to_snapshot(self) -> T:
```
Tomas2D: this method does not exist in TypeScript

python/beeai_framework/serialization/serializer.py
```diff
@@ -0,0 +1,313 @@
...
+# Copyright 2025 © BeeAI a Series of LF Projects, LLC
+# SPDX-License-Identifier: Apache-2.0
+import json
+from collections.abc import Callable, Iterable, Iterator, Mapping
+from dataclasses import dataclass
+from datetime import datetime
+from itertools import count
+from typing import Any, ClassVar, Generic, TypeVar
+
+T = TypeVar("T")
```
Tomas2D, line 13: Would be possible to set a default to a `dict`? Otherwise, I would remove it completely.

python/beeai_framework/serialization/serializer.py
Sakeeb91: I've updated both create_snapshot and load_snapshot to be abstract methods with Awaitable return types to match the TypeScript implementation:
@abstractmethod
def create_snapshot(self) -> T | Awaitable[T]:
    raise NotImplementedError

@abstractmethod
def load_snapshot(self, snapshot: T) -> None | Awaitable[None]:
    raise NotImplementedError
The Serializable class now inherits from ABC, and both methods use the @abstractmethod decorator. This ensures type safety and aligns with the TypeScript interface at [serializable.ts:25-26](https://github.com/i-am-bee/beeai-framework/blob/main/typescript/src/internals/serializable.ts#L25-L26) where both methods can return Promise types.

python/beeai_framework/serialization/serializer.py
Sakeeb91: I've removed the to_snapshot method as it doesn't exist in the TypeScript implementation. Looking at the TypeScript code, users call createSnapshot() directly rather than having a convenience wrapper. The Python implementation now matches this pattern.

python/beeai_framework/serialization/serializer.py
Sakeeb91: I investigated adding a default value to the `T` TypeVar. The `default` parameter for `TypeVar` is only available in Python 3.13+, but this project supports Python 3.11-3.13 (as specified in `pyproject.toml`). Following your suggestion to "remove it completely" if a default isn't possible, I've kept the TypeVar without a default:
`T = TypeVar("T")`
This maintains compatibility across all supported Python versions while keeping the generic type parameter flexible for different snapshot types.

python/beeai_framework/serialization/serializer.py
```diff
@@ -0,0 +1,408 @@
...
+        if class_name == "Object":
+            result: dict[str, Any] = {}
+            if not isinstance(payload, Mapping):
+                raise SerializerError("Encountered malformed payload for JavaScript Object.")
+            remember(result)
+            for key, item in payload.items():
+                result[key] = cls._decode(item, seen)
+            return result
+
+        return cls._NOT_HANDLED
```
Tomas2D: Built-ins should be also registered as every other data types as it is done in TypeScript. No messy if statements.

python/beeai_framework/backend/message.py
```diff
@@ -374,3 +374,89 @@ def dedupe_tool_calls(msg: AssistantMessage) -> None:
...
+        "tool-call": MessageToolCallContent,
+    },
+)
+
+_register_message_class(
+    ToolMessage,
+    {
+        "tool-result": MessageToolResultContent,
+    },
+)
```
Tomas2D: This might not be needed if we make an arbitrary Pydantic model a serializable which should be easy.

```python
Seriazer.register(BaseModel, to_plain=lambda value: value.model_dump(), from_plain=lambda value, ref: ref.model_validate(value)) 
```

Note: the current implementation of `from_plain` differs from the one defined in TS. In TS the second parameter is a reference to the factory (class/function). It is not implemented here.

Tomas2D (changes requested):
Good work so far! Thank you!

Tomas2D:
Hello @Sakeeb91. Would you be willing to address the comments I mentioned? I am ready to help if you need me. Thank you!

Sakeeb91:
@Tomas2D Thank you for the buzz. I may have overlooked your previous comments and gone on with my life. I have addressed the changes. I hope they are alright!

python/beeai_framework/backend/message.py
Tomas2D: The message parts should inherit from `Serializable` instead of adding these functions. Please update.

Tomas2D (changes requested):
My comments have not been addressed. Please take a look at my comments regarding serialization of the message parts.

Sakeeb91:
Addressed the review feedback. Message content classes now inherit from `Serializable` via a new `SerializableModel` base class that combines `BaseModel` with `Serializable`. The `Message` class also inherits from `Serializable` directly. Removed the `_register_message_class` function and all manual registration calls.

All serialization and backend tests pass.

Tomas2D:
@Sakeeb91 following up — you noted on 31 Dec that the review feedback was addressed, but there's been no activity since and the branch now conflicts with `main` (`mergeable: CONFLICTING`).

To move forward this needs a rebase onto current `main`, and then a confirmation that the message-parts serialisation comments are resolved against the current code, since `main` has moved substantially in eight months.

Still happy to take this if you want to pick it back up. If not, say the word and we'll close it — reopening is always possible. Absent a reply I'll close it as stale in a couple of weeks.

Tomas2D:
Status check on this one — it's been open since October 2025 and has drifted badly against `main`.

**The docs target no longer exists.** This PR edits `docs/modules/serialization.mdx`, but docs have since migrated to Astro Starlight. The live file is now [docs/src/content/docs/modules/serialization.mdx](https://github.com/i-am-bee/beeai-framework/blob/main/docs/src/content/docs/modules/serialization.mdx) (the old tree survives only as `docs-old/`). That's a guaranteed conflict independent of the code changes, and the surrounding format has changed too.

**Python serialization doesn't exist on `main`.** `python/beeai_framework/serialization/` is untracked — only the TypeScript side has it (`typescript/examples/serialization/*.ts`). So this isn't "add examples" as the title suggests; it's introducing the whole Python serialization subsystem plus examples plus docs, and it touches `backend/message.py`, `memory/token_memory.py` and `memory/unconstrained_memory.py` to do it.

That's a substantial feature, and it already carries a changes-requested review and 11 comments of feedback. Given eleven months of drift on files that are actively maintained, a rebase is likely to be more work than it's worth.

@Sakeeb91 — are you still planning to pick this up? If so, it'd be worth confirming with a maintainer that Python serialization is still wanted in this shape before you invest in the rebase. If not, closing this and opening a fresh PR against current `main` would be cleaner for everyone.

PR #1397, opened 2026-01-27
feat(tools): add scratchpad tool for agent working memory

ezelanza:
## Summary

This PR adds a new scratchpad tool that provides agents with a persistent working memory to track their reasoning, actions, and observations. This helps agents avoid repeating actions and maintain context across interactions.

## Features

- **Read operation**: View current scratchpad contents
- **Write operation**: Add or update entries with key-value pair merging
- **Append operation**: Append text to the last entry
- **Clear operation**: Reset the scratchpad
- **Session-based storage**: Maintains persistent storage across interactions
- **Key-value pair merging**: Automatically consolidates entries to prevent duplicates

## Implementation Details

- Follows existing code style and conventions
- Includes proper copyright headers (Apache 2.0)
- Full type annotations with modern Python syntax
- Passes all linting checks (Black, Ruff)
- Comprehensive test suite with 15 unit tests covering:
  - Basic operations (read, write, append, clear)
  - Key-value pair handling
  - Error handling
  - Input validation
  - Session management

## Testing

All tests pass locally. The implementation includes:
- Input validation tests
- Operation tests for all four operations
- Edge case handling (empty scratchpad, missing content, etc.)
- Session management tests

## Checklist

- [x] Code follows project style guidelines
- [x] Copyright headers added
- [x] Type annotations included
- [x] Linting passes (Black, Ruff)
- [x] Unit tests added
- [x] Conventional commit format used
- [x] DCO sign-off included


python/beeai_framework/tools/scratchpad/scratchpad.py
```diff
@@ -0,0 +1,343 @@
...
+        handler = handlers.get(operation)
+        if handler:
+            result = handler()
+            return StringToolOutput(result=result)
+
+        error_msg = (
+            f"Unknown operation: {operation}. "
+            "Use 'read', 'write', 'append', or 'clear'."
+        )
+        return StringToolOutput(result=error_msg)
```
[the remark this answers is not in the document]
ezelanza: Solved

python/tests/tools/test_scratchpad.py
```diff
@@ -0,0 +1,207 @@
...
+
+    read_result = await tool.run(input={"operation": "read"})
+    assert "Dict entry" in read_result.result
+
+
+@pytest.mark.asyncio
+async def test_missing_operation_field(tool: ScratchpadTool) -> None:
+    """Test that missing operation field raises validation error."""
+    with pytest.raises(ToolInputValidationError):
+        await tool.run(input={})
```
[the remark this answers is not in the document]
ezelanza: Solved

python/tests/tools/test_scratchpad.py
```diff
@@ -0,0 +1,207 @@
...
+    await tool.run(input=ScratchpadInput(operation="write", content="city: Boston"))
+
+    # Write second set of pairs (should merge)
+    await tool.run(input=ScratchpadInput(operation="write", content="date: 2025-01-28"))
+
+    # Read it back
+    result = await tool.run(input=ScratchpadInput(operation="read"))
+    assert "city: Boston" in result.result
+    assert "date: 2025-01-28" in result.result
+
```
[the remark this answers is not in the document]
ezelanza: Solved

python/beeai_framework/tools/scratchpad/scratchpad.py
```diff
@@ -0,0 +1,343 @@
...
+        pairs = {}
+        # Split by comma, but be careful with commas inside values
+        parts = [p.strip() for p in content.split(",")]
+        for part in parts:
+            if ":" in part:
+                key, value = part.split(":", 1)
+                key = key.strip()
+                value = value.strip()
+                if key and value:
+                    pairs[key] = value
```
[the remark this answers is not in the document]
ezelanza: Solved

sandijean90:
Thank you for adding this and contributing @ezelanza ! I'm glad it helped you with your use case and you're sharing it with other BeeAI users.

@Tomas2D can you check out this scratchpad implementation and provide your feedback before merging?

python/beeai_framework/tools/scratchpad/scratchpad.py
```diff
@@ -0,0 +1,456 @@
...
+from beeai_framework.context import RunContext
+from beeai_framework.emitter import Emitter
+from beeai_framework.tools import (
+    StringToolOutput,
+    Tool,
+    ToolInputValidationError,
+    ToolRunOptions,
+)
+
+logger = logging.getLogger(__name__)
```
Tomas2D: Use BeeAI logger instead.

```python
from beeai_framework.logger import Logger

logger = Logger(__name__)
```

python/beeai_framework/tools/scratchpad/scratchpad.py
```diff
@@ -0,0 +1,456 @@
...
+        return (
+            "Manage your working memory (scratchpad). Use this to track "
+            "what you've done, what results you got, and avoid repeating "
+            "actions. Operations: 'read' - see your scratchpad, 'write' - "
+            "add an entry, 'clear' - reset scratchpad, 'append' - add to "
+            "existing entry."
+        )
+
+    @property
+    def input_schema(self) -> type[BaseModel]:
```
Tomas2D: `type[ScratchpadInput]:`

python/beeai_framework/tools/scratchpad/scratchpad.py
```diff
@@ -0,0 +1,456 @@
...
+        """Input schema for the tool."""
+        return ScratchpadInput
+
+    def _create_emitter(self) -> Emitter:
+        """Create emitter for the tool."""
+        return Emitter()
```
Tomas2D: This would create a completely new event tree. 

Do this instead.

```python
Emitter.root.child(
  namespace=["tool", "scratchpad"], 
  creator=self
)
````

File: python/beeai_framework/tools/scratchpad/scratchpad.py [hunk 8]
```diff
@@ -0,0 +1,456 @@
...
+        logger.info(
+            f"ScratchpadTool[{session_id}]: CLEAR - " f"{entries_count} entries"
+        )
+        return result
+
+    async def _run(
+        self,
+        input: ScratchpadInput,
+        options: ToolRunOptions | None = None,
+        context: RunContext | None = None,
```
Tomas2D: `context` is always defined (in addition it can be retrieved via `RunContext.get()` anywhere in the code

python/beeai_framework/tools/scratchpad/scratchpad.py
```diff
@@ -0,0 +1,456 @@
...
+        # Try different context attributes in order of preference
+        session_id = None
+
+        # run_id: Should persist across tool calls in the same agent run
+        if hasattr(context, "run_id") and context.run_id:
+            session_id = str(context.run_id)
+            logger.debug(f"Using run_id as session: {session_id}")
+
+        # conversation_id: If available, persists across the conversation
+        elif hasattr(context, "conversation_id") and context.conversation_id:
```
Tomas2D: This attribute does not exist on the `RunContext`.

python/beeai_framework/tools/scratchpad/scratchpad.py
```diff
@@ -0,0 +1,456 @@
...
+            session_id = str(context.conversation_id)
+            logger.debug(f"Using conversation_id as session: {session_id}")
+
+        # agent_id: If available, unique per agent instance
+        elif hasattr(context, "agent_id") and context.agent_id:
```
Tomas2D: This attribute does not exist on the `RunContext`.

Tomas2D (commented):
Thank @ezelanza for your contribution. 

Would you mind adding some practical examples showcasing the tool in action? Probably in some long-running process or in a multi-agent setup, as I don't currently see a motivation for `scratchpads` to be bound to the class as one can share the tool instance among multiple agents during initialization. In addition, some attributes don't even exist on the `RunContext`. Please see my comments.

ezelanza:
> Thank @ezelanza for your contribution.
> 
> Would you mind adding some practical examples showcasing the tool in action? Probably in some long-running process or a multi-agent setup, since I don't currently see a motivation for `scratchpads` to be bound to the class; one can share the tool instance among multiple agents during initialization. In addition, some attributes don't even exist on the `RunContext`. Please see my comments.

Thanks @Tomas2D! I've found Scratchpad to solve a problem in my scenario where I have multiple agents behind a supervisor, and I'm using a relatively Small local model (8B), which most of the times get confused to follow the rules of the prompt (a 200-line one), and forcing it (the supervisor) to read the scratchpad was a game-changer. I didn't use it to share info between agents; I've used it to improve an agent's understanding and memory in that scenario.

Will update the fields you highlighted

ezelanza:
@Tomas2D I've updated the code based on your comments. Could you take a look and let me know if there is anything else to change or modify?

Thanks!

Tomas2D:
Hello @ezelanza . I don't see any changes relevant to my latest comment, which addresses the problems mentioned.

Tomas2D:
ping @ezelanza

Tomas2D:
@ezelanza checking in — this has been waiting on the same feedback since February, with pings on 12 Feb and 6 Apr and no response since.

To restate what's still open, so picking it up is easy:
- The motivation for binding scratchpads to the class instance is still unclear, given a single tool instance can be shared across agents at initialisation. The multi-agent / long-running example asked for would settle it.
- Some attributes the session-key logic reads don't exist on `RunContext`.

The branch is also ~7 months behind `main` and will need a rebase.

If you're still interested, comment here and take your time — no rush from our side. If you'd rather not continue, that's completely fine too; just let us know and we'll close it so it stops showing up as pending. Absent a reply I'll close it as stale in a couple of weeks, and it can always be reopened.

Tomas2D (changes requested):
This has sat since January and CI is red (the run logs have since expired, so I reviewed the diff directly). Several things need addressing before it can land.

**Lint — almost certainly the CI failure.** `ruff` has `I` (isort) enabled in `python/pyproject.toml`, and the imports are:

```python
import asyncio
from beeai_framework.logger import Logger   # ← first-party import between two stdlib imports
import re
```

That's an `I001` failure. `mise fix` will sort it.

**Doesn't follow the `Tool` generic convention.** `class ScratchpadTool(Tool)` is unparameterized; every other tool in the repo declares its types, e.g. [openmeteo.py:45](https://github.com/i-am-bee/beeai-framework/blob/main/python/beeai_framework/tools/weather/openmeteo.py#L45):

```python
class OpenMeteoTool(Tool[OpenMeteoToolInput, ToolRunOptions, JSONToolOutput[dict[str, Any]]]):
```

Should be `Tool[ScratchpadInput, ToolRunOptions, StringToolOutput]`. As written, `pyrefly` has nothing to check the run signature against.

**`operation` isn't actually validated.**

```python
operation: str = Field(..., enum=["read", "write", "append", "clear"])
```

`enum=` is not a pydantic v2 `Field` parameter — it's silently swallowed into `json_schema_extra`, so `operation="drop_table"` passes validation and only fails later in your dispatch. The repo convention is a `Literal`, as in `temperature_unit: Literal["celsius", "fahrenheit"]` in openmeteo.py. Use `operation: Literal["read", "write", "append", "clear"]` and you get validation, a correct JSON schema for the model, and the enum in the tool description for free.

**Process-global mutable state is the real design question.** `_scratchpads: ClassVar[dict[str, list[str]]]` is shared by every `ScratchpadTool` instance in the process, for the process's lifetime. Your docstring acknowledges the unbounded growth and pushes `clear_session()` onto consumers, but that's an easy contract to break — an agent run that raises never cleans up, and the framework gives you no hook that guarantees it. Two agents in one process also share a namespace and can collide if the API layer sets the same `session_id`.

The framework already has per-run state that solves this: `RunContext` is the natural owner, or instance-level storage with the tool constructed per agent. I'd push back on the class-level dict rather than document around it — a memory leak in a library is much harder for users to diagnose than in their own code.

Also `_lock: ClassVar[asyncio.Lock]` is instantiated at class-definition (import) time. It's safe on 3.10+ since locks no longer bind a loop at construction, but a single process-wide lock shared across every event loop is fragile, and it disappears entirely if the storage becomes per-run.

**Minor:** `"Content to write/append (required for 'write' and 'append' " "operations)"` is an implicit concat that renders as `"...'append' operations)"` — fine, but it's what `ruff`'s `ISC` rules exist to catch; just make it one string.

Happy to re-review once the lint and the `Literal` are sorted, but the storage-lifetime question is worth a maintainer's opinion first, since it may change the shape of the whole tool.

PR #1450, opened 2026-05-13
fix: preserve LiteLLM text with tool calls

### LINKED ISSUE #1448: The `_transform_output` method of `ChatModel` may discard text content or override reasoning content with possible side effect of AttributeError
**Describe the bug**
The `_transform_output` method uses simple conditional to build `AssistantMessage`. If any tool calls are made, the text content of the message, if there is one, is dropped.

At the same time, the `AssistantMessage` is constructed using either `content` or `reasoning_content` field of `Delta`, in this order of precedence. This means that if both are defined, the reasoning content is ignored.  

```
AssistantMessage(update.content or update.reasoning_content or "", id=chunk.id)
```
However, both fields can be populated at the same time. [1] In fact, it is possible to have `tool_calls`, `content` and `reasoning_content` set on the same `Delta` object.

But the initialization of `AssistantMessage` introduces a third problem. If the `Delta` object from litellm is initialized without any reasoning content, the attribute `reasoning_content` is deleted. 

```
        if reasoning_content is not None:
            self.reasoning_content = reasoning_content
        else:
            # ensure default response matches OpenAI spec
            del self.reasoning_content
```

This is a problem introduced by litellm. But it will possibly manifest here, if the `content` field doesn't evaluate to true.
In that case the the attempt to access `reasoning_content` will fail with `AttributeError`.

In principle, that situation is rather unlikely, as it would require the inference provider not to set reasoning, or regular content.
But it can still happen. I have seen it happen with some Gemini models running in GCP. 

**To Reproduce**
1. Create agent with `LiteLLMChatModel`
2. Start the agent and compare raw inference output to responses handled by agent.

**Expected behavior**
All content received from inference should be processed into `AssistantMessage` objects.

**Set-up:**

- Bee version: 0.1.80
- Model provider Google Gemini

**Additional context**
Litellm source:

[1] https://github.com/BerriAI/litellm/blob/aa587bd9d3c1f21decd590a68fd07427f1fa3653/litellm/types/utils.py#L1263

asim48-ctrl:
## Summary
- preserve LiteLLM assistant text content when tool calls are present
- avoid AttributeError when LiteLLM omits `reasoning_content` on a delta/message
- add focused regression coverage for text+tool-call and missing-reasoning cases

Fixes #1448.

## Verification
- `uv run --no-sync ruff check beeai_framework/adapters/litellm/chat.py tests/backend/providers/test_litellm_chat.py`
- `uv run --no-sync ruff format beeai_framework/adapters/litellm/chat.py tests/backend/providers/test_litellm_chat.py --check`
- `uv run --no-sync pytest tests/backend/providers/test_litellm_chat.py -q`
- `git diff --check`

### DIFF
--- a/python/beeai_framework/adapters/litellm/chat.py
+++ b/python/beeai_framework/adapters/litellm/chat.py
@@ -310,23 +310,28 @@ class LiteLLMChatModel(ChatModel, ABC):
 
         output: list[AnyMessage] = []
         if update and update.model_dump(exclude_none=True):
-            content: list[MessageTextContent | MessageToolCallContent] = []
             text = update.content if update.content is not None else getattr(update, "reasoning_content", None)
-            if text is not None:
-                content.append(MessageTextContent(text=text))
-
             tool_calls = getattr(update, "tool_calls", None)
-            if tool_calls:
-                content.extend(
-                    MessageToolCallContent(
-                        id=call.id or "",
-                        tool_name=call.function.name or "",
-                        args=call.function.arguments,
-                    )
-                    for call in tool_calls
-                )
+            if text is not None:
+                text_id = f"{chunk.id}:text" if tool_calls else chunk.id
+                output.append(AssistantMessage(MessageTextContent(text=text), id=text_id))
 
-            output.append(AssistantMessage(content, id=chunk.id))
+            if tool_calls:
+                output.append(
+                    AssistantMessage(
+                        [
+                            MessageToolCallContent(
+                                id=call.id or "",
+                                tool_name=call.function.name or "",
+                                args=call.function.arguments,
+                            )
+                            for call in tool_calls
+                        ],
+                        id=chunk.id,
+                    )
+                )
+            elif text is None:
+                output.append(AssistantMessage("", id=chunk.id))
 
         return ChatModelOutput(
             output=output,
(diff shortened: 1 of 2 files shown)


Tomas2D (commented):
I would say that this is breaking change. Can the bug be fixed without merging it together?

asim48-ctrl:
Thanks, agreed. I pushed a follow-up that no longer merges text and tool calls into the same `AssistantMessage`.

The updated shape preserves LiteLLM text as a separate output chunk when tool calls are also present, while keeping the tool-call message shape separate. It still fixes the missing `reasoning_content` AttributeError path.

Re-verified locally:
- `uv run --no-sync ruff check beeai_framework/adapters/litellm/chat.py tests/backend/providers/test_litellm_chat.py`
- `uv run --no-sync ruff format beeai_framework/adapters/litellm/chat.py tests/backend/providers/test_litellm_chat.py --check`
- `uv run --no-sync pytest tests/backend/providers/test_litellm_chat.py -q`
- `git diff --check`

Tomas2D (commented):
What is the difference after your changes?

asim48-ctrl:
The updated version avoids the breaking shape I had in the first pass.

Before this PR, when LiteLLM returned both `content` and `tool_calls`, BeeAI only emitted the tool-call message and dropped the text content.

After the latest changes:
- text-only responses still produce one text `AssistantMessage`
- tool-call-only responses still produce one tool-call `AssistantMessage`
- text + tool-call responses produce two output items: first the text `AssistantMessage`, then the existing tool-call `AssistantMessage` shape
- missing `reasoning_content` no longer raises `AttributeError`

So the tool-call message shape is kept separate; the only behavioral addition is preserving text that LiteLLM already returned instead of discarding it.

PR #1352, opened 2025-12-07
DeepEval Tutorial 

yatishdurga:
### Which issue(s) does this pull-request address?

Closes: #

#### General
- [ ] I have read the appropriate contributor guide: [Python](https://github.com/i-am-bee/beeai-framework/blob/main/CONTRIBUTING.md)
/ [TypeScript](https://github.com/i-am-bee/beeai-framework/blob/main/CONTRIBUTING.md)
- [ ] I have signed off on my commit: [Python instructions](https://github.com/i-am-bee/beeai-framework/blob/main/python/CONTRIBUTING.md#developer-certificate-of-origin-dco) / [TypeScript instructions](https://github.com/i-am-bee/beeai-framework/blob/main/typescript/CONTRIBUTING.md#developer-certificate-of-origin-dco)
- [ ] Commit messages and PR title follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary)
- [ ] Appropriate label(s) added to PR: `Python` for Python changes, `TypeScript` for TypeScript changes

#### Code quality checks
- [ ] Code quality checks pass: `mise check` (`mise fix` to auto-fix)

#### Testing
- [ ] Unit tests pass: `mise test:unit`
- [ ] E2E tests pass: `mise test:e2e`
- [ ] Tests are included (for bug fixes or new features)

#### Documentation
- [x] Documentation is updated
- [ ] Embedme embeds code examples in docs. To update after edits, run: Python `mise docs:fix`

PR #1453, opened 2026-05-15
feat(python): add evaluation framework (DeepEval + Ragas)

RoyiAizenstain:

Closes: #<issue-number>

### Description

Adds a Python **evaluation framework** for systematically measuring BeeAI
agent quality on multi-hop QA, tool usage, and factual accuracy. Rather
than eyeballing outputs, users can run a fixed dataset of questions
through an agent and score the responses with pluggable metrics.

**What's included:**

- **Adapters** (`python/evaluation/adapters/`) that bridge BeeAI's
  `ChatModel` to two popular eval frameworks, so any of the 26+ BeeAI
  providers can serve as the judge LLM:
  - `DeepEvalLLM`  → `deepeval.DeepEvalBaseLLM`
  - `InstructorRagasLLM` → `ragas.InstructorBaseRagasLLM`
- **Reference experiments** (`python/examples/evaluation/`):
  - Shared `agent.py` (a `RequirementAgent` for multi-hop QA with
    Wikipedia / OpenMeteo / PythonTool) and `dataset.json` ground truth.
  - `deepeval/experiment.py` — runs the agent against the dataset and
    scores it with built-in DeepEval metrics plus three custom ones
    (`AnswerLLMJudgeMetric`, `ToolUsageMetric`, `FactsSimilarityMetric`).
  - `ragas/experiment.py` — async-script runner with built-in Ragas
    metrics plus a Pydantic-structured `FactsSimilarityMetric`.
- **Docs** (`docs/modules/evaluation.mdx`) — adapter usage, env vars,
  invocation, output file locations.

**Design highlights:**

- Adapter pattern keeps eval frameworks pluggable behind BeeAI's
  `ChatModel`.
- Agent-under-test and judge-LLM are decoupled via `AGENT_CHAT_MODEL_NAME`
  / `EVAL_CHAT_MODEL_NAME` env vars to avoid self-bias.
- Default fallback model is `ollama:llama3.1:8b`, matching BeeAI's
  conventional default.
- No module-load side effects: `ragas/dataset.py` exposes
  `build_dataset()` callable, no pytest decorators in example scripts.

**No breaking changes** — purely additive; existing modules untouched
apart from new docs.

#### General
- [x] I have signed off on my commit: [Python instructions](https://github.com/i-am-bee/beeai-framework/blob/main/python/CONTRIBUTING.md#developer-certificate-of-origin-dco)
- [x] Appropriate label(s) added to PR: `Python`

#### Code quality checks

#### Testing

#### Documentation
- [x] Documentation is updated
- [x] Embedme embeds code examples in docs. To update after edits, run: Python `mise docs:fix`


RoyiAizenstain:
can you review it @Tomas2D ?
Thanks :)

docs/modules/evaluation.mdx
```diff
@@ -0,0 +1,159 @@
...
+
+## Dataset & Agent Setup
+
+Evaluation experiments require two components: a **dataset** of test cases and an **agent** to evaluate.
+
+### Dataset
+
+The dataset is a JSON file where each item contains:
+
+| Field                  | Description                                            |
```
Tomas2D: Small mismatch: this table documents `expected_answer`, `supporting_sentences`, and `expected_tool_calls`, but `python/examples/evaluation/dataset.json` actually uses `answer`, `relevant_sentences`, and `wiki_times`. Could you align the two? (Honestly, the names documented here are the clearer ones — renaming the JSON fields to match might be the nicer fix.)

docs/src/content/docs/modules/evaluation.mdx
```diff
@@ -0,0 +1,159 @@
...
+## Adapters
+
+The adapters bridge BeeAI's `ChatModel` to each evaluation framework's LLM interface, so metrics can use BeeAI-powered models for judging agent outputs without requiring separate API configurations.
+
+### DeepEvalLLM
+
+Wraps a BeeAI `ChatModel` as a `DeepEvalBaseLLM`, enabling DeepEval metrics to call any BeeAI-supported model for LLM-as-judge evaluations. Supports structured output via Pydantic schemas.
+
+```py Python [expandable]
+from evaluation.adapters import DeepEvalLLM
```
Tomas2D: Heads-up: this presents `evaluation.adapters` as framework API, but `python/evaluation/` is a repo-level directory that isn't part of the published `beeai_framework` wheel — so users who `pip install beeai-framework` won't be able to run this import. If we want the adapters to be public API, they should live under `beeai_framework/` (perhaps gated behind an optional extra); otherwise the docs should say explicitly that this requires a repo checkout.

python/evaluation/adapters/deepeval_llm.py
```diff
@@ -43,21 +40,28 @@ async def a_generate(self, prompt: str, schema: TSchema | None = None) -> str:
...
+        options: ModelLike[ChatModelParameters] | None = None,
+        **kwargs: Any,
     ) -> "DeepEvalLLM":
         name = name or KEY_FILE_HANDLER.fetch_data(ModelKeyValues.LOCAL_MODEL_NAME)
-        # pyrefly: ignore [bad-argument-type]
+        if isinstance(name, str) and name.startswith("vertexai:"):
+            merged: dict = {"allow_prompt_caching": False}
+            if isinstance(options, dict):
+                merged.update(options)
+            options = merged or options
```
Tomas2D: Two thoughts here:

1. `merged` is always non-empty at this point, so `merged or options` always evaluates to `merged` — the `or options` part is dead code.
2. A `vertexai:`-specific workaround hardcoded inside a generic adapter (and duplicated in `examples/evaluation/agent.py`) feels like the wrong layer. If prompt caching misbehaves on VertexAI, the fix probably belongs in the VertexAI backend adapter — or at minimum only in the example, with a comment explaining why.

python/evaluation/adapters/deepeval_llm.py
```diff
@@ -29,11 +28,9 @@ def __init__(self, model: ChatModel, *args: Any, **kwargs: Any) -> None:
     def load_model(self, *args: Any, **kwargs: Any) -> None:
         return None
 
-    # pyrefly: ignore [bad-override]
     def generate(self, prompt: str, schema: BaseModel | None = None) -> str:
```
Tomas2D, line 39: The removed `# pyrefly: ignore [bad-override]` comments were added during the pyrefly migration (#1347), and the overridden signatures haven't changed — so I'd expect `mise check` to fail without them. Could you run `mise check` and restore whatever it needs?

python/evaluation/adapters/ragas_llm.py
```diff
@@ -0,0 +1,70 @@
...
+
+logger = logging.getLogger(__name__)
+
+T = TypeVar("T", bound=BaseModel)
+
+
+class InstructorRagasLLM(InstructorBaseRagasLLM):
+    """A class that bridges Ragas with BeeAI directly (without LangChain intermediary)."""
+
+    def __init__(self, model_name: str):
```
Tomas2D: For consistency with `DeepEvalLLM` (which wraps an existing `ChatModel`), consider `__init__(self, model: ChatModel)` plus a `from_name()` factory. Taking only a name string means callers can't pass a configured model (options, parameters, middleware).

Relatedly, `from_name(model_name, **kwargs)` below forwards `**kwargs` into this constructor, which accepts none — so any kwarg raises `TypeError`.

python/evaluation/adapters/ragas_llm.py
```diff
@@ -0,0 +1,70 @@
...
+        1. Takes a Ragas request.
+        2. Converts it to a format BeeAI understands (UserMessage).
+        3. Uses BeeAI's native response_format for structured output.
+        4. Returns a Pydantic object.
+        """
+        native_message = UserMessage(prompt)
+        response = await self.model.run([native_message], response_format=response_model)
+        raw_text = response.get_text_content()
+
+        clean_json = raw_text.strip()
```
Tomas2D: Since the call above already passes `response_format=response_model`, the backend should be returning valid JSON — if markdown fences still show up, that's likely a backend/provider bug worth surfacing rather than papering over here. If the stripping turns out to be necessary for some providers, a short comment naming which ones would help future readers.

python/beeai_framework/evaluation/adapters/ragas_llm.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def generate(self, prompt: str, response_model: Type[T]) -> T:
+        """Synchronous version (required to implement due to inheritance)."""
+        try:
+            loop = asyncio.get_running_loop()
+        except RuntimeError:
+            loop = None
+
+        if loop and loop.is_running():
+            import nest_asyncio
```
Tomas2D: `nest_asyncio` isn't declared in `pyproject.toml`, so this path raises `ModuleNotFoundError` for anyone hitting the sync API inside a running loop. Either declare it (optional group) or drop the sync fallback and document the adapter as async-only — both seem fine.

python/examples/evaluation/deepeval/experiment.py
```diff
@@ -0,0 +1,219 @@
...
+    for msg in messages:
+        if isinstance(msg, ToolMessage):
+            for item in msg.content:
+                tool_name = getattr(item, "tool_name", None)
+                if tool_name and tool_name != "final_answer":
+                    tool_counter[tool_name] += 1
+    return dict(tool_counter)
+
+
+async def create_rag_test_cases():
```
Tomas2D: There's already a shared harness in `evaluation/_utils.py` (renamed in this PR): `create_dataset()` builds test cases with `EVAL_CACHE_DATASET` caching, and `evaluate_dataset()` prints rich pass/fail tables. This file re-implements both (including the hand-rolled ASCII table further down and the defensive `getattr(eval_results, ...)` probing). Could you reuse the shared helpers — or extend them if they're missing something you need? That keeps the example focused on what's actually new.

python/examples/evaluation/deepeval/experiment.py
```diff
@@ -0,0 +1,219 @@
...
+load_dotenv()
+
+logger = logging.getLogger(__name__)
+
+from examples.evaluation.agent import create_agent
+from examples.evaluation.dataset import load_items
+from evaluation.adapters import DeepEvalLLM
+from beeai_framework.backend import ToolMessage
+
+from AnswerLLMJudgeMetric import AnswerLLMJudgeMetric
```
Tomas2D: These bare module imports only resolve when the CWD is the script directory, and the PascalCase module names (`AnswerLLMJudgeMetric.py`, etc.) will trip ruff's N999 check. Suggest snake_case module names (`answer_llm_judge_metric.py`) and package-relative imports.

python/examples/evaluation/deepeval/answer_llm_judge_metric.py
```diff
@@ -0,0 +1,64 @@
...
+            score = float(str(text).strip())
+        except (ValueError, TypeError):
+            score = 0.0
+
+        score = max(0.0, min(1.0, score))
+        self.score = score
+        self.success = score >= self.threshold
+        return score
+
+    def measure(self, test_case: LLMTestCase) -> float:
```
Tomas2D, line 58: `asyncio.run()` raises `RuntimeError` when called from a running event loop — which is the common case under `deepeval.evaluate` with async metrics. Since `async_mode = True` already routes everything through `a_measure`, consider having `measure()` raise `NotImplementedError` (like the adapter does) rather than appearing to work. Same applies to `FactsSimilarityMetric.measure()`.

python/examples/evaluation/ragas/experiment.py
```diff
@@ -0,0 +1,197 @@
...
+from ragas.metrics.collections import AnswerAccuracy, ContextPrecision, ContextRecall, ExactMatch, ToolCallAccuracy
+
+from examples.evaluation.agent import create_agent
+from evaluation.adapters import InstructorRagasLLM
+from FactsSimilarityMetric import FactsSimilarityMetric
+
+# Load dataset
+_script_dir = Path(__file__).parent
+_data_dir = _script_dir / "data"
+dataset = Dataset.load(name="my_evaluation", backend="local/csv", root_dir=str(_data_dir))
```
Tomas2D: The PR description mentions "no module-load side effects", but this module loads the dataset and constructs the judge LLM at import time — so importing it fails outright if `dataset.py` hasn't been run yet. Moving both into `main()` would make the example match the description (and play nicer with tooling that imports modules to inspect them).

python/examples/evaluation/ragas/experiment.py
```diff
@@ -0,0 +1,197 @@
...
+logger = logging.getLogger(__name__)
+
+# Add python/ root to path for shared evaluation module
+sys.path.insert(0, str(Path(__file__).parent.parent.parent.parent))
+# Add this folder to path for local metric imports
+sys.path.insert(0, str(Path(__file__).parent))
+
+# Suppress multiprocess resource tracker warnings on Windows - must be before other imports
+# TODO: remove once multiprocess fixes ResourceTracker.__del__ on Windows
+if sys.platform == "win32":
```
Tomas2D, line 21: This is quite a lot of workaround machinery for an example users will copy as a starting point. The `sys.path.insert` calls above can go away with proper package-relative imports, and for the `ResourceTracker` monkey-patch it would be good to link the upstream `multiprocess` issue it works around — or drop it if it only suppresses noise at interpreter shutdown.

python/examples/evaluation/agent.py
```diff
@@ -0,0 +1,82 @@
...
+    "if the question requires it.",
+    "3. FINAL FORMAT: Your ONLY final output MUST be a single, valid JSON object adhering "
+    "strictly to the required keys: answer, tool_used, supporting_titles, supporting_sentences, "
+    "reasoning_explanation. The final_answer must be concise and specific (e.g., just 'Delhi', "
+    "not a full sentence). Do not include any text outside the JSON block.",
+    "4. THE JSON SCHEMA STRING: " + JSON_SCHEMA_STRING,
+]
+
+
+def create_calculator_tool() -> PythonTool:
```
Tomas2D: The example agent unconditionally includes `PythonTool`, which needs a code interpreter running at `http://127.0.0.1:50081` — but neither dataset question requires it, and the docs never mention starting one. Suggest dropping it from the default agent or making it opt-in via an env var, so the example works out of the box.

python/examples/evaluation/agent.py
```diff
@@ -0,0 +1,82 @@
...
+                    'ollama:llama3.1:8b'.
+    """
+    if model_name is None:
+        model_name = os.environ.get(
+            "AGENT_CHAT_MODEL_NAME",
+            os.environ.get("EVAL_CHAT_MODEL_NAME", "ollama:llama3.1:8b"),
+        )
+
+    options: dict = {"allow_parallel_tool_calls": True}
+    if model_name.startswith("vertexai:"):
```
Tomas2D: Same `vertexai:` workaround as in `DeepEvalLLM.from_name` — see my comment there; it would be nice to have this live in one place (ideally the VertexAI backend) rather than duplicated.

Tomas2D (changes requested):
Thanks so much for this contribution, @RoyiAizenstain! 🐝 The Ragas adapter and the runnable end-to-end examples are a genuinely valuable addition — having reference experiments people can copy and adapt is exactly what the evaluation story needs. I left inline comments throughout; here I'll summarize the structural points that need attention before we can merge:

1. **Docs need a rebase onto the new docs system.** While this PR was open, the docs migrated from Mintlify to Astro Starlight (#1462), so `docs/docs.json` and `docs/modules/` no longer exist on `main` and the branch is currently in merge conflict. Could you rebase and move the new page to `docs/src/content/docs/modules/evaluation.mdx` using the Starlight frontmatter/components? Happy to help if the new setup is unclear.

2. **PR description vs. what's new.** `DeepEvalLLM` already exists at `python/eval/model.py` (added in #1352) — this PR moves it rather than adds it. The genuinely new pieces (the Ragas adapter and the examples) are great; it would just help reviewers if the description reflected that.

3. **Stranded tutorial.** The `eval/` → `evaluation/` rename moves every file except `python/eval/deep_eval tutorial.mdx`, which would be left behind as the only file in `python/eval/` — and it overlaps with the new docs page. Could you move/merge or remove it as part of this PR?

4. **Missing dependency declarations.** The new code imports `ragas`, `nest_asyncio`, `numpy`, and `multiprocess`, but none of them are declared in `python/pyproject.toml` (only `deepeval = "3.3.5"` exists today). Could you add them with pins — ideally behind an optional `evaluation` group — and double-check that the DeepEval symbols used (`ExactMatchMetric`, `ArgumentCorrectnessMetric`, `DEEPEVAL_PER_TASK_TIMEOUT_SECONDS_OVERRIDE`) exist in the pinned version?

5. **Checklist.** DCO sign-off, `mise check`, and the test runs are still unchecked, and `Closes: #<issue-number>` is a leftover placeholder. A few of the new files are also missing trailing newlines (`agent.py`, `dataset.py`, `dataset.json`) and carry `Copyright 2025` headers — `mise fix` should take care of most of this.

None of this diminishes the value of the work — once it's rebased and aligned with the existing harness, this will be a really nice addition. Thank you again! 🙏

Tomas2D:
Hi @RoyiAizenstain — circling back on this one. Whenever you have time, the main open items from my review are:

- Rebasing onto `main` (the branch currently conflicts after the docs migration from Mintlify to Astro Starlight), and moving the docs page under `docs/src/content/docs/modules/`.
- Signing off the commits so DCO passes.
- The code points in the inline comments.

No rush, and happy to help if any of it is unclear. Thanks again for the contribution!

RoyiAizenstain:
Thanks! I will review and fix what you asked for :) @Tomas2D

docs/modules/evaluation.mdx
RoyiAizenstain: Done — renamed the JSON fields to match your suggestion rather than the other direction: answer→expected_answer, relevant_sentences→supporting_sentences, wiki_times→expected_tool_calls in dataset.json, updated all consumers (deepeval/experiment.py, ragas/dataset.py) and the docs table to match.

python/evaluation/adapters/ragas_llm.py
RoyiAizenstain: You were right to flag this — response.output_structured was already returning validated output via the framework's own json_repair-based parser, so the manual fence-stripping was solving an already-solved problem. Removed it entirely; agenerate() now just uses response.output_structured directly and raises ChatModelError if the backend fails to produce it.

python/examples/evaluation/deepeval/answer_llm_judge_metric.py
RoyiAizenstain: Fixed both — AnswerLLMJudgeMetric.measure() and FactsSimilarityMetric.measure() now both raise NotImplementedError instead of the unsafe asyncio.run() bridge (I'd initially only caught the first one on re-review, both are fixed now).

python/examples/evaluation/ragas/experiment.py
RoyiAizenstain: Removed the sys.path.insert calls (switched to proper package-relative imports) and dropped the ResourceTracker monkey-patch entirely, per your second option — confirmed empirically that the underlying AttributeError at shutdown doesn't affect the exit code (Python suppresses __del__ exceptions by design), and the patch itself turned out to be broken at late shutdown anyway. Also dropped a related atexit/multiprocessing.util._exit_function() block for the same reason. No upstream issue link added — searched uqfoundation/multiprocess's tracker and found nothing filed matching this exact symptom.

docs/src/content/docs/modules/evaluation.mdx
RoyiAizenstain: Went with making it real public API — moved the adapters into beeai_framework/evaluation/adapters/, added a new evaluation extra to pyproject.toml (pip install "beeai-framework[evaluation]"), updated the docs accordingly.

python/evaluation/adapters/deepeval_llm.py
RoyiAizenstain: Restored — ran pyrefly check (confirmed it genuinely fails without these, exactly as you predicted) and put back all 4 # pyrefly: ignore [bad-override] comments (load_model, generate, a_generate, get_model_name). File's now clean under pyrefly again.

RoyiAizenstain:
@Tomas2D All 14 inline comments are addressed. A few highlights beyond the direct asks:

- Moved `evaluation.adapters` into `beeai_framework/` behind a new `evaluation` extra (`pip install "beeai-framework[evaluation]"`), so it ships in the wheel instead of requiring a checkout
- `ragas_llm.py`'s `agenerate()` now uses `response.output_structured` directly instead of manual JSON extraction — turned out the framework already handles that
- Renamed `dataset.json`'s fields to your suggested names (`expected_answer`, `supporting_sentences`, `expected_tool_calls`)
- Dropped the `ResourceTracker` monkey-patch and `atexit` workaround entirely after confirming they don't affect anything (verified the shutdown error doesn't touch the exit code)
- Found and fixed a real bug along the way: `ArgumentCorrectnessMetric` was always scoring 0 because tool call args were never being captured


Ready for re-review whenever you have time. Thanks for the thorough feedback.

Tomas2D (commented):
Thanks for working through all 14 comments, @RoyiAizenstain — and for moving `evaluation.adapters` into `beeai_framework/` behind an `evaluation` extra, that was the right call.

I approved the gated CI run. `Lint & Build & Test` is **red** on two pyrefly errors, both in `examples/evaluation/deepeval/experiment.py`:

```
ERROR Could not import `ExactMatchMetric` from `deepeval.metrics` [missing-module-attribute]
  --> examples/evaluation/deepeval/experiment.py:21:5

ERROR Unexpected keyword argument `model` in function
      ToolCorrectnessMetric.__init__ [unexpected-keyword]
  --> examples/evaluation/deepeval/experiment.py:163:31
163 |         ToolCorrectnessMetric(model=eval_model, include_reason=False),
```

Both look like the example being written against a different DeepEval version than the one the lockfile resolves: `ExactMatchMetric` isn't exported from `deepeval.metrics`, and `ToolCorrectnessMetric` is a non-LLM metric so it takes no `model`. Worth pinning a lower bound on `deepeval` in the `evaluation` extra once you've settled on the version, so this can't drift again — an example that doesn't import is worse than no example.

I'll do the full re-review pass over the substantive changes once CI is green; flagging these now so they're not blocking on me. If either error is a pyrefly false positive rather than a real API mismatch, say so and I'll look again.

RoyiAizenstain:
Thanks! Both errors were real, deepeval was pinned exact at 3.3.5, which predates ExactMatchMetric (added in 3.7.0) and predates ToolCorrectnessMetric accepting model at all.

Bumped the pin to ^3.7.0 (resolves to 3.9.9), regenerated poetry.lock. Verified both classes match experiment.py as-is - no code changes needed beyond the version bump.

Tomas2D (commented):
Re-reviewed at `fb732a46`. The pin bump did fix both errors I reported — and immediately surfaced a third, which is the one blocking thing here. The rest of this is the substantive pass I owed you.

## Blocking: CI is still red

```
ERROR Could not import `LLMTestCaseParams` from `deepeval.test_case` [missing-module-attribute]
  --> evaluation/agents/requirement/test_greeting.py:9:32
9 | from deepeval.test_case import LLMTestCaseParams
```

This one isn't your code — the import is pre-existing (it's in `python/eval/agents/requirement/test_greeting.py` on `main`) and this PR only carries it through the `eval/` → `evaluation/` rename. But `main` pins `deepeval = "3.3.5"` exactly, so nothing had exercised it against a newer release until now. Going to `^3.7.0` (→ 3.9.9) moved or removed that symbol.

Since you have the environment resolved, please find where `LLMTestCaseParams` lives in 3.9.x and update the import and its five usages in the `GEval(evaluation_params=[...])` call. Worth grepping the whole `evaluation/` tree for other 3.3-era imports at the same time — this is unlikely to be the only one, and pyrefly stops at the first error.

## Packaging

**`numpy`, `pandas`, `scipy`, `multiprocess` shouldn't be direct dependencies.** None of them appear anywhere in `main`'s dependency list, and none of the code in this PR imports them — they're transitive requirements of `deepeval` and `ragas`. Declaring them directly means we pin and own version ranges for the scientific stack forever, and we'll be the ones fielding the resolver conflicts. Let deepeval and ragas bring their own; drop all four from both `evaluation` and `all`. `nest-asyncio` is a genuine direct dependency (you import it) so that one stays.

## `InstructorRagasLLM.generate()`

```python
if loop and loop.is_running():
    import nest_asyncio
    nest_asyncio.apply()
    return loop.run_until_complete(self.agenerate(prompt, response_model))
```

`nest_asyncio.apply()` patches asyncio **globally, for the whole process**, as a side effect of someone calling a sync method on one object. An application that imports us for something unrelated gets a re-entrant event loop it never asked for, and this is close to impossible to trace back when it misbehaves later.

I'd rather not ship that from library code. Options, roughly in order of preference: raise a clear error telling the caller to use `agenerate()` when a loop is already running; or run the coroutine on a dedicated worker thread with its own loop (`asyncio.run_coroutine_threadsafe` / a small `ThreadPoolExecutor`); or, if Ragas genuinely gives us no way out, keep the patch but move it behind an explicit opt-in and document it loudly. The `asyncio.run()` path for the no-loop case is fine as-is.

## Unrelated changes to split out

Two things in here have nothing to do with evaluation, and both should be their own PR so they're individually reviewable and revertable:

- **`.gitignore` adds `CLAUDE.md`.** That's personal tooling — it belongs in your global gitignore or `.git/info/exclude`, not in the shared repo. The `.deepeval/` and eval-output entries below it are legitimately part of this PR.
- **`adapters/vertexai/backend/chat.py`** now forces `allow_prompt_caching = False` by default. That's a real fix with a real upstream reference (BerriAI/litellm#17696), but it silently changes behaviour for every VertexAI user and shouldn't ride in on an evaluation PR. Send it separately and I'll take it quickly.

## What's good

Worth saying: moving `evaluation.adapters` into `beeai_framework/` behind the extra was the right call and is done properly. The `ModuleNotFoundError` → "install `beeai-framework[evaluation]`" guard is the right pattern and matches the other optional adapters. `agenerate()` correctly reuses the framework's native `response_format` instead of reimplementing JSON repair, and raising `ChatModelError` with the raw output in `context` on a schema mismatch is exactly right. The docs page is solid.

Fix the import, drop the four transitive deps, resolve the `nest_asyncio` question, and split the two unrelated changes — then I think this is ready.

RoyiAizenstain:
Thanks for the thorough pass!

- Fixed `LLMTestCaseParams` → `SingleTurnParams`, ran live end-to-end to confirm.
- Dropped numpy/pandas/scipy/multiprocess as direct deps.
- `generate()` now raises a clear error on a running loop instead of patching asyncio; also dropped `nest-asyncio` since it's no longer imported.
- Split out `.gitignore`'s `CLAUDE.md` line and the VertexAI caching default. The latter is up as its own PR: #1626 <link once opened>.

All pushed to `eval-pr` at `1a7dc657`.

RoyiAizenstain:
This work was done under the guidance of @antonpibm and in collaboration with @OmriMeroz 💪🏻

Special thanks to @Tomas2D 😀

PR #1455, opened 2026-05-17
fix(python): ignore empty tool messages

pragnyanramtha:
## Summary
- avoid indexing the first tool result while scanning memory for the last tool call/response pair
- use the last tool call from a multi-tool assistant message when selecting the matching tool response
- ignore empty or unrelated tool messages instead of raising `IndexError`
- add regression coverage for matching tool responses, empty tool messages, and multi-tool assistant messages

## Validation
- `uv run --with pytest --with syrupy pytest tests/memory/test_utils.py -q` -> 3 passed; emitted local warnings for an unknown asyncio config option and PyCharm diff patching
- `uv run --with ruff ruff check beeai_framework/memory/utils.py tests/memory/test_utils.py`
- `uv run --with ruff ruff format --check beeai_framework/memory/utils.py tests/memory/test_utils.py`
- `git diff --check`

## Notes
The commits include DCO signoffs as required by the Python contributing guide.

### DIFF
--- a/python/beeai_framework/memory/utils.py
+++ b/python/beeai_framework/memory/utils.py
@@ -18,11 +18,13 @@ def extract_last_tool_call_pair(memory: BaseMemory) -> tuple[AssistantMessage, T
         return None
 
     tool_call: AssistantMessage = memory.messages[tool_call_index]  # type: ignore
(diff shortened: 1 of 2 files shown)


Tomas2D (approved):
thank you!

PR #1483, opened 2026-06-11
fix(tools): clean up LLMTool prompt template

### LINKED ISSUE #1475: Bug: garbled prompt template in LLMTool
The prompt template in `typescript/src/tools/llm.ts` (~line 58) contains merged sentence fragments sent to the model on every call: "You have to accomplish a task by using Using common sense and the information contained in the conversation up to this point, complete the following task." Rewrite it into a coherent sentence.

77652189:
## Summary
- rewrite the default LLMTool task prompt into a coherent sentence
- add a regression test for the rendered default prompt

Closes #1475

## Testing
- `corepack yarn vitest run src/tools/llm.test.ts`
- `corepack yarn prettier --log-level warn --check src/tools/llm.ts src/tools/llm.test.ts`
- `corepack yarn eslint src/tools/llm.ts src/tools/llm.test.ts`

### DIFF
--- a/typescript/src/tools/llm.ts
+++ b/typescript/src/tools/llm.ts
@@ -55,7 +55,7 @@ export class LLMTool extends Tool<StringToolOutput, LLMToolInput> {
     schema: z.object({
       task: z.string(),
     }),
-    template: `You have to accomplish a task by using Using common sense and the information contained in the conversation up to this point, complete the following task. Do not follow any previously used formats or structures.
+    template: `Use common sense and the information contained in the conversation up to this point to complete the following task. Do not follow any previously used formats or structures.
 
 The Task: {{task}}`,
   });
--- a/typescript/src/tools/llm.test.ts
+++ b/typescript/src/tools/llm.test.ts
@@ -1,0 +1,19 @@
+/**
+ * Copyright 2025 BeeAI a Series of LF Projects, LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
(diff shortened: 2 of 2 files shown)


typescript/src/tools/llm.ts
```diff
@@ -55,7 +55,7 @@ export class LLMTool extends Tool<StringToolOutput, LLMToolInput> {
     schema: z.object({
       task: z.string(),
     }),
-    template: `You have to accomplish a task by using Using common sense and the information contained in the conversation up to this point, complete the following task. Do not follow any previously used formats or structures.
```
Tomas2D, line 58: does it correspond to the python version?

77652189: Thanks for checking. I looked through the Python package and could not find a direct counterpart for `LLMTool` or this default prompt template under `python/`. This change only updates the TypeScript `LLMTool` prompt. If there is a Python equivalent I missed, I would be happy to align the wording there as well.

Tomas2D: Thanks for checking, @77652189 — I confirmed there's no `LLMTool` (or equivalent prompt template) under `python/`, so a TypeScript-only change is correct here. The fix reads much better than the original "...by using Using common sense...", and the added test is a nice touch.

Tomas2D (approved):
All green and the change is correct (verified there's no Python counterpart for LLMTool). Approving — thank you for the contribution! 🐝

Issue #1471, opened 2026-06-10
Update vulnerable and outdated dependencies


Tomas2D:
Audit dependencies in both packages and update any with known vulnerabilities or security advisories:

- Python: `poetry show --outdated` / `pip-audit` in `python/`
- TypeScript: `yarn audit` / `yarn outdated` in `typescript/`

Bump affected packages, verify tests still pass, and open a PR per package ecosystem.

Tomas2D:
Logging a specific advisory to pick up here (consolidated from #1494):

- **basic-ftp** — [GHSA-5rq4-664w-9x2c](https://github.com/advisories/GHSA-5rq4-664w-9x2c) (CVE-2026-27699), path traversal in `downloadToDir()`, affects `< 5.2.0`. We currently resolve **5.0.5** transitively via `pac-proxy-agent → get-uri → basic-ftp`. Fix: add a `resolutions` entry (`"basic-ftp": ">=5.2.0"`) in `typescript/package.json` and regenerate `yarn.lock`. Low real-world risk (no FTP usage in source), but worth clearing the alert.

For the record, the PyTorch RCE report (#1495) was **not** applicable — we already ship `torch` 2.7.1, past the 2.6.0 fix for CVE-2025-32434.

weijun-xia:
Can I take the TS side of this? main already resolves basic-ftp to 5.3.1, so I'll add the `resolutions` pin (`basic-ftp: >=5.2.0`) to lock the minimum and clear the advisory.

Tomas2D:
Status update — both ecosystems have been audited and remediated:

- **Python** ✅ merged in #1511 — `pip-audit` 47 → 3. Residuals (`torch` CVE-2025-3000, `diskcache` CVE-2025-69872, `pytest` CVE-2025-71176) have no upstream patch or are dev-only; none reachable from framework code.
- **TypeScript** 🔵 #1515 (open, mergeable) — `yarn npm audit` 41 → 28 via `resolutions` pins, including the **basic-ftp** lock (CVE-2026-27699) called out above. Residuals are dev-tooling / optional integrations / deprecated leaves with no safe fix; the `vitest` (critical) item is a Vitest **UI dev-server** advisory not reachable from headless `vitest run` and is deferred to a dedicated vitest 2→3 migration. None of the residuals ship in the published package's runtime dependencies.

#1515 carries `Closes #1471`, so this issue will close automatically once it is merged.

PR #1500, opened 2026-06-17
fix(deps): pin basic-ftp to ^5.2.0 to clear GHSA-5rq4-664w-9x2c

weijun-xia:
## What
Pin `basic-ftp` to `^5.2.0` via the `resolutions` field in `typescript/package.json` and regenerate `yarn.lock`.

## Why
`basic-ftp` <5.2.0 is affected by GHSA-5rq4-664w-9x2c (CVE-2026-27699), a path traversal in `downloadToDir()`. It is pulled in transitively via `pac-proxy-agent -> get-uri -> basic-ftp`, as noted in #1471.

## Change
- `typescript/package.json`: add `"basic-ftp": "^5.2.0"` to `resolutions`.
- `typescript/yarn.lock`: the `basic-ftp` descriptor moves from `^5.0.2` to `^5.2.0` and resolves to **5.3.1** (already past the 5.2.0 fix), so there is no version churn and no added packages.

## Note on the range
The issue suggested `>=5.2.0`. I used `^5.2.0` instead, because `>=5.2.0` resolves to `basic-ftp@6.0.1` (a major bump, +7 packages / ~71 MiB) which `get-uri` (`^5.0.2`) does not declare support for. `^5.2.0` clears the advisory while staying within the compatible major. Happy to switch to `>=5.2.0` if you prefer.

## Verification
- `yarn install --mode=update-lockfile` regenerates the lockfile with `basic-ftp@5.3.1`.
- `yarn npm audit --all --recursive` no longer lists the basic-ftp advisory.
- The diff is limited to the two files above; `basic-ftp` is not used in source, so there is no runtime impact.

Other unrelated advisories remain in some dev dependencies (e.g. `@zilliz/milvus2-sdk-node`, `@octokit/plugin-paginate-rest`); those are out of scope here and can be a follow-up.

Ref: #1471


Tomas2D:
Superseded by #1515 (merged), which pins `basic-ftp` via `resolutions` at `^5.3.1` (stronger than `^5.2.0`) together with the other transitive advisories. Thanks for flagging it @xiaweiwei67-stack!

Issue #1479, opened 2026-06-10
Add OpenAI-compatible server (OpenAIServer) to TypeScript


Tomas2D:
Python has an OpenAI-compatible server (`python/beeai_framework/adapters/openai/serve` — chat completions + responses endpoints) that exposes any agent as a standard OpenAI API. TypeScript only has `a2a` and `mcp` serve adapters. There is real demand for this: Watsonx Orchestrate consumes the chat completions endpoint (see discussion #1008, where the Python implementation originated).

**Dependency:** maintainers wanted new server adapters to build on the unified Runnable interface (discussion #971). Python has `runnable.py`; TypeScript does not have a Runnable interface yet — porting it is likely a prerequisite (or this issue should be scoped to whatever interim agent abstraction maintainers prefer).

Reference implementation: `python/beeai_framework/adapters/openai/serve/server.py`.

SahebSandhuSingh:
Hey @Tomas2D , I would like to work on this, if anyone else has not been appointed.
I have gone through the codebase
Happy to open a draft PR early for feedback.

Veselin-Vasilev-IBM-1:
Hi @Tomas2D , I would also like to work on that. Thanks.

PR #1502, opened 2026-06-17
feat(infra): port Runnable interface to TypeScript

SahebSandhuSingh:
## Summary

Ports Python's `Runnable` interface (introduced in #971) to TypeScript. This is a prerequisite for the OpenAI-compatible serve adapter requested in #1479, since the issue notes that new server adapters should be built on the unified `Runnable` interface, and TypeScript currently has no equivalent.

## What's included

- **`Runnable<R>`** — abstract class mirroring Python's `Runnable[R]` ABC, with `run()`, `emitter`, and `middlewares`
- **`RunnableOutput`** — matches Python's `RunnableOutput`, including the `lastMessage` getter with its empty-`AssistantMessage` fallback
- **`RunnableOptions`** — `{ signal?, context? }`, matching Python's `RunnableOptions` TypedDict
- **`runnableEntry`** — a helper function that wraps `RunContext.enter()`, used as the TypeScript equivalent of Python's `runnable_entry` decorator

## Design notes / deviations from Python

- **Decorator → helper function**: Python uses a `@runnable_entry` decorator to auto-wrap `run()`. TS decorators are still experimental/stage-3 and behave inconsistently across build tools, and none of the existing serve adapters in this codebase (`A2AServer`, `MCPServer`) use decorators anywhere. A plain helper function (`runnableEntry`) was used instead, to match the existing code style.
- **`extra="allow"` on `RunnableOutput`**: Python's pydantic model allows arbitrary extra fields via `extra="allow"`. TS doesn't have a direct equivalent, so an index signature (`[key: string]: any`) was used as an approximation. Open to feedback on whether there's a more idiomatic pattern already used elsewhere in this codebase for this.
- **Input mutation via middleware**: Python's `runnable_entry` reads the (possibly middleware-modified) input from `ctx.run_params` rather than the original argument, so middleware can rewrite input before the handler runs. `runnableEntry` replicates this by reading `context.runParams` inside the wrapped handler — covered by a regression test.

## Testing

- `runnable.test.ts` covers:
  - Basic `run()` flow and output shape
  - `lastMessage` fallback to an empty `AssistantMessage` when output is empty
  - Middleware mutating input before the handler runs (regression test for the behavior above)
- `tsc --noEmit` passes clean for both files
- `eslint` passes clean for both files

## Status

Opening as a **draft PR** for early feedback on direction and naming before going further — happy to adjust based on maintainer input, especially on the `RunnableOutput` extra-fields approach and whether `runnableEntry`'s signature/naming fits expectations.


Tomas2D (changes requested):
Thanks @SahebSandhuSingh — the core of this is sound: `runnable.ts` is clean, purely additive, correctly targets the real `RunContext.enter(...)` API, and the middleware regression test genuinely exercises the intended behavior. Two things must be cleaned up before merge:

- **Remove the committed `package-lock.json` (~19.9k lines).** This project uses **Yarn 4 (Berry)** (`.yarnrc.yml`, `packageManager: "yarn@4.9.1"`); an npm lockfile must not be committed and will cause dependency drift.
- **Revert the unrelated `yarn.lock` churn** (~2.5k lines of dependency bumps like `@ai-sdk/anthropic`, `@a2a-js/sdk`, new `@ai-sdk/deepseek`). This looks like an accidental `npm install` — please regenerate with `yarn install` so the diff is just the two new files.

Also: the interface isn't exported from `src/index.ts`, so it's currently unreachable to consumers — please add the export (or note it's intentionally internal for now). Once the lockfiles are reverted and the fork-gated CI passes, the source itself looks good.

SahebSandhuSingh:
Ok @Tomas2D I will do the needful

SahebSandhuSingh:
@Tomas2D Addressed all three:

- Removed package-lock.json
- Reset yarn.lock to match upstream/main and reinstalled — diff is clean, no unrelated dependency churn
- Exported Runnable from src/index.ts

Shall i start with OpenAI-compatible server adapter for TypeScript??

Tomas2D:
Thanks @SahebSandhuSingh — confirmed the cleanup: the `package-lock.json` is gone, the `yarn.lock` churn is reverted (diff is just the three intended files), and Runnable is now exported from `index.ts`. 

CI is red on **commitlint**: the commit header is 116 chars (limit is 100). Please shorten it (e.g. `feat(infra): port Runnable interface to TypeScript`) and force-push — should go green after that, and I'll approve.

Re your question: no need to start a separate OpenAI-compatible TS server — that's already being built in #1507. This Runnable port is the useful groundwork for it.

SahebSandhuSingh:
@Tomas2D this is also done, kindly check

Tomas2D:
Merged — thank you @SahebSandhuSingh! 🎉 Really faithful port of the `Runnable` interface: the `RunnableOutput` extra-fields handling, the generic message types, and especially the `runnableEntry` helper reading from `context.runParams` (so middleware mutations take effect, mirroring Python's `runnable_entry`) all show great attention to the reference implementation. The only thing blocking it was Prettier formatting on the two new files, which I applied for you before merging. Excellent contribution! 🐝

PR #1547, opened 2026-07-01
feat(memory): add DakeraMemory for persistent cross-session agent memory

### LINKED ISSUE #1546: Add DakeraMemory for persistent cross-session agent memory
## Problem

BeeAI agents start each session with a blank context. There is no built-in `BaseMemory` implementation that persists across process restarts or chat sessions. Long-running assistants, support bots, or research agents need to recall facts from prior conversations without passing the entire history every time.

## Proposed solution

A new `DakeraMemory` class implementing `BaseMemory` that backs every conversation turn with a self-hosted [Dakera](https://dakera.ai) memory server.

Dakera provides:
- **On-device semantic search** via an ONNX embedding model — no cloud dependency
- **Decay-weighted importance** — frequently accessed memories stay relevant; stale facts fade
- **Session scoping** — memories can be namespaced per session

### How it works

```typescript
import { DakeraMemory } from "beeai-framework/memory/dakeraMemory";
import { ReActAgent } from "beeai-framework/agents/react/agent";
import { OllamaChatModel } from "beeai-framework/adapters/ollama/backend/chat";

const memory = new DakeraMemory({
  url: "http://localhost:3300",
  apiKey: "demo",        // optional — omit for local dev
  agentId: "bee-agent",
  sessionId: "sess-123", // optional — scopes recall to one session
  topK: 5,               // past memories injected per turn
  persist: true,         // false = read-only (replay without polluting store)
});

const agent = new ReActAgent({
  llm: new OllamaChatModel("granite4:micro"),
  tools: [],
  memory,
});

// Run the agent — DakeraMemory handles store + recall automatically
const response = await agent.run({ prompt: "What did I ask you last time?" });
```

**Run Dakera locally:**
```bash
docker run -p 3300:3300 -e DAKERA_API_KEY=demo ghcr.io/dakera-ai/dakera:latest
```

### Design

- `add(message)`: persists to `POST /v1/memory/store`, then fetches top-K relevant past memories via `POST /v1/memory/search` and injects them as system messages into the current window
- Non-fatal failures: Dakera outages emit `console.warn` and the agent continues normally
- `snapshot` / `loadSnapshot` round-trip correctly for BeeAI checkpoint serialization

PR #XXX contains the implementation with 8 unit tests (all HTTP-mocked).

ferhimedamine:
## Summary

Adds `DakeraMemory` — a `BaseMemory` implementation that gives BeeAI agents persistent, semantically-searchable cross-session memory backed by a self-hosted [Dakera](https://dakera.ai) server.

Closes #1546

## What this PR adds

**`typescript/src/memory/dakeraMemory.ts`** — full `BaseMemory` implementation:
- `add(message)` → persists to Dakera, enriches the window with top-K recalled memories
- `delete(message)` → removes from the in-session list
- `reset()` → clears the in-session window (Dakera store unaffected)
- `recall(query, topK?)` / `forgetSession()` → explicit recall + per-session cleanup helpers
- `createSnapshot` / `loadSnapshot` — BeeAI serializer support (the API key is **never** serialized)
- `static { this.register() }` — enrolled with the framework serializer (same pattern as `TokenMemory`, `SlidingMemory`)

**`typescript/src/memory/dakeraMemory.test.ts`** — 10 Vitest unit tests, all HTTP-mocked (no live server).

## How it works

```typescript
import { DakeraMemory } from "beeai-framework/memory/dakeraMemory";
import { ReActAgent } from "beeai-framework/agents/react/agent";

const memory = new DakeraMemory({
  url: "http://localhost:3000",
  apiKey: "dk-...",      // optional
  agentId: "bee-agent",
  topK: 5,               // past memories to recall per turn
});

const agent = new ReActAgent({ llm, tools: [], memory });
await agent.run({ prompt: "What did I ask you last time?" });
```

**Run Dakera locally** with the public [`dakera-ai/dakera-deploy`](https://github.com/dakera-ai/dakera-deploy) docker-compose (the server needs the object store the compose provisions — a bare `docker run` of the image is not enough):

```bash
git clone https://github.com/dakera-ai/dakera-deploy
cd dakera-deploy && docker compose up -d   # server listens on :3000
```

## API mapping

| DakeraMemory | Dakera endpoint |
|---|---|
| `add()` (persist=true) | `POST /v1/memory/store` |
| `add()` recall injection / `recall()` | `POST /v1/memory/recall` |
| `forgetSession()` | `POST /v1/memory/forget` |

## Design decisions

- **`persist: false`** — read-only mode (recalls but never writes). Useful for replaying prior conversations without polluting the store.
- **API key never serialized** — `createSnapshot()` omits `apiKey`; on deserialization it is dropped and must be re-supplied for write access.
- **Non-fatal outages** — Dakera store/recall failures are logged via the framework `Logger` (warn) and the agent continues; the in-session window always retains the current turn.
- **Recalled memories as a single system message** — injected with `meta: { __dakera_injected: true }` so it can be filtered; prior injections are purged before fresh ones (prevents unbounded growth), including on a no-hit recall.

## Testing

This is a **TypeScript-only** change (no Python).

- [x] 10 Vitest unit tests pass (`yarn vitest run src/memory/dakeraMemory.test.ts`)
- [x] `yarn tsc --noEmit`, `yarn eslint`, `yarn prettier --check` all clean
- [x] `yarn tsup` build emits `dist/memory/dakeraMemory.{js,cjs,d.ts}` — importable via the `beeai-framework/memory/*` subpath export, like every built-in memory
- [x] Store failure (HTTP 500) is non-fatal — agent continues
- [x] `serialize()` → `fromSerialized()` round-trip verified with `verifyDeserialization`; a secret-leak regression test asserts the API key never appears in the checkpoint

## Review follow-ups (this revision)

Addresses @Tomas2D's review:
- **Test runner** — rewritten to Vitest globals + `verifyDeserialization` from `@tests/e2e/utils.js`; dropped `@jest/globals`.
- **🔴 Secret leak** — `apiKey` removed from `createSnapshot()`/`loadSnapshot()`; added a regression test.
- **Convention drift** — `console.warn` → framework `Logger`; the class builds to `dist/` and is importable via the standard `memory/*` subpath.
- **Endpoint** — recall now maps to `POST /v1/memory/recall` (the decay-weighted agent recall endpoint) instead of `/v1/memory/search`.
- **Docs** — corrected the port to `:3000` and pointed setup at `dakera-deploy`; clarified the change is TypeScript-only.

Also folded in the Gemini review findings (read-only injection, purge-on-no-hit, `createdAt` type guard, test placeholders).


ferhimedamine:
Thanks for the thorough review! All four issues have been addressed in the latest commit:

**1. `persist: false` now correctly allows recall injection** ✅
The early-return guard `if (!this.persist || ...)` was incorrectly skipping recall injection in read-only mode. Fixed by splitting the guards: the Dakera store is skipped when `persist=false`, but the injection step always runs (read-only mode still benefits from long-term recall).

**2. Memory leak: `purgeInjected()` now runs before the search** ✅
`purgeInjected()` was inside `if (hits.length > 0)`, so a no-hit search left stale injected messages in the window. Fixed: always purge first, then only insert if there are new hits.

**3. `createdAt` TypeError guard** ✅
Added `instanceof Date` check before calling `.toISOString()` — the value could be a string when restored from a snapshot.

**4. Test constructor empty strings** ✅
Replaced `new DakeraMemory({ url: "", agentId: "" })` (which throws `MemoryError`) with valid placeholder values, then used `loadSnapshot()` to overlay the real snapshot for verification.

Tomas2D (changes requested):
Thanks for this, @ferhimedamine — a persistent cross-session memory adapter is a nice idea. A few blockers before it can land:

- **Tests target the wrong runner.** The test file imports `{ jest } from "@jest/globals"` and uses `jest.fn()`, but this repo runs **Vitest** (`vitest ^2.1.9`); `@jest/globals` isn't a dependency. Existing memories (`slidingMemory.test.ts`, `tokenMemory.test.ts`) use Vitest globals plus `verifyDeserialization` from `@tests/e2e/utils.js`. As written these tests won't compile/run here.
- **🔴 Secret leak: the API key is serialized.** `createSnapshot()` includes `apiKey: this.apiKey`, so any persisted/logged checkpoint contains the key in cleartext. Please exclude `apiKey` from the snapshot.
- **Convention drift.** Uses `console.warn` for error handling instead of the framework `Logger`, which also silently degrades to "no memory" if the server is unreachable. And the class isn't wired into the package exports, so the import shown in the PR body may not resolve.
- **Likely wrong endpoint.** Recall maps to `POST /v1/memory/search`; Dakera's docs describe `POST /v1/memory/recall` — worth verifying against the actual API.
- The fork-gated CI never ran, so none of the above was caught by the build. Also, the PR body says "Python" but the change is TypeScript-only.

The concept is welcome — happy to see a revised version.

ferhimedamine:
Thanks for the thorough review, @Tomas2D — all five blockers are addressed in `5d1b4a6`. Point by point:

**1. Tests target the wrong runner.** Rewritten to Vitest: dropped `@jest/globals`/`jest.fn()` in favour of the Vitest globals (`vi`, `describe`, `it`) and added a `serialize()` → `fromSerialized()` round-trip using `verifyDeserialization` from `@tests/e2e/utils.js`, matching `tokenMemory.test.ts`/`slidingMemory.test.ts`. `yarn vitest run src/memory/dakeraMemory.test.ts` → **10/10 pass**.

**2. 🔴 Secret leak — API key serialized.** `apiKey` is now excluded from `createSnapshot()`. Because deserialization bypasses the constructor (`Object.create`), `loadSnapshot()` explicitly drops it (the key must be re-supplied for write access after a restore). Added a regression test that asserts the key appears in **neither** the snapshot object nor the serialized checkpoint string.

**3. Convention drift.** `console.warn` → the framework `Logger` (a module-scoped `Logger.root.child({ name: "DakeraMemory" })`), so outages now surface as structured `warn` records in the framework log stream rather than silently. On the exports: like every built-in memory, the class is picked up by the `src/**/*.ts` tsup entry and resolves through the `./*` → `./dist/*` subpath export — I confirmed `yarn tsup` emits `dist/memory/dakeraMemory.{js,cjs}`, so `import { DakeraMemory } from "beeai-framework/memory/dakeraMemory"` resolves (I corrected the import path in the PR body to match).

**4. Wrong endpoint.** Correct — recall now maps to `POST /v1/memory/recall` (Dakera's decay-weighted, agent-scoped recall: `{ query, agent_id, top_k }` → `{ memories: [{ memory, score }] }`), not `/v1/memory/search`. The public helper is renamed `recall()` to match.

**5. Fork-gated CI + "Python" wording.** The change is TypeScript-only — I removed the stray wording and marked it explicitly in the PR body. Since the fork workflow needs your approval to run, I ran the exact CI steps locally against this repo and they're green:
- `yarn tsc --noEmit` ✅
- `yarn eslint` ✅
- `yarn prettier --log-level warn --check .` ✅
- `yarn tsup` build ✅ (emits the dist artifacts above)
- `yarn vitest run src/memory/dakeraMemory.test.ts` ✅ 10/10

I also folded in the Gemini findings (read-only injection, purge-before-recall on a no-hit response, `createdAt` type guard, test placeholders). Happy to iterate further — thanks again.

ferhimedamine:
Following up on your endpoint question, @Tomas2D — I verified the wire contract against the Dakera server source directly (not just the docs):

- `POST /v1/memory/store` → `{ memory, embedding_time_ms }`
- `POST /v1/memory/recall` → `{ memories: [{ memory, score, … }], … }`
- `POST /v1/memory/forget` → `{ deleted_count }`

These are the exact routes the server registers, and the recall response is the nested `{ memory, score }` shape the adapter maps (`hit.memory.content`) — so the `/v1/memory/recall` mapping is confirmed correct (the `/search` variant is a separate keyword-search route we don't use here).

All five blockers from your review remain addressed in `5d1b4a6` (Vitest tests, `apiKey` excluded from the snapshot, framework `Logger`, subpath export via `beeai-framework/memory/dakeraMemory`, TypeScript-only). Whenever you have a moment for another pass, it should be ready — happy to rebase if you'd like the branch brought up to `main` first.

ferhimedamine:
Rebased onto `main` (currently at #1552) — the branch is now a single clean commit `6a3d053` with no merge commit, so the history is linear again. The diff is unchanged (the two `dakeraMemory` files); all the earlier review fixes are intact. Ready for another look, @Tomas2D 🙏

vivek41-glitch:
Use git commit with sign-off  and dco will pass

Tomas2D:
Thanks @ferhimedamine — nice rework: the tests are on Vitest now, and `apiKey` is correctly excluded from `createSnapshot()` with a test asserting it. 

CI is red on **DCO/commitlint**: _"message must be signed off"_. Please sign off your commit(s) — `git commit --amend -s` (or `git rebase --signoff` for several) then force-push.

One higher-level point for the maintainers before this can land: it adds a first-party adapter in core `src/memory/` for Dakera, which is currently a public-alpha / open-core third-party product. Whether that belongs in core vs. a community/contrib location is a project-policy call worth confirming — flagging for @Tomas2D. Happy to keep reviewing once the sign-off is in.

ferhimedamine:
> Thanks @ferhimedamine — nice rework: the tests are on Vitest now, and `apiKey` is correctly excluded from `createSnapshot()` with a test asserting it. 
> 
> CI is red on **DCO/commitlint**: _"message must be signed off"_. Please sign off your commit(s) — `git commit --amend -s` (or `git rebase --signoff` for several) then force-push.
> 
> One higher-level point for the maintainers before this can land: it adds a first-party adapter in core `src/memory/` for Dakera, which is currently a public-alpha / open-core third-party product. Whether that belongs in core vs. a community/contrib location is a project-policy call worth confirming — flagging for @Tomas2D. Happy to keep reviewing once the sign-off is in.
> 

Thanks @Tomas2D for review and feedback. I will address git signing issue.

ferhimedamine:
Done — signed off the commit as requested (`git commit --amend -s`) and force-pushed. New head: `ddd576f4`.

- `Signed-off-by: Mohamed Amine Ferhi <ferhi.med.amine@gmail.com>` now matches the commit author, so both the **DCO** check and commitlint's **`signed-off-by`** rule are satisfied.
- Only the commit message changed — the diff (the two `dakeraMemory` files) is byte-for-byte identical to before.

One heads-up: the **TypeScript – Lint, Build, Test** workflow runs as `pull_request` from a fork, so it needs a maintainer *approve-and-run* on this new commit before it executes. Whenever you get a moment to approve it, CI should go green.

On the core-vs-contrib placement you flagged: happy to relocate the adapter to a community/contrib path if the team prefers — just say the word. Thanks again for the review!

ferhimedamine:
Heads-up: this PR was **accidentally auto-closed** when the branch was force-pushed with a commit that lost its shared history with `main` (a shallow-clone mistake on my side). GitHub won't allow reopening a PR whose recorded head has no common ancestor with the base branch, so I've restored the branch history, signed off the commit (resolving the DCO/commitlint red), and continued the exact same work here:

👉 **https://github.com/i-am-bee/beeai-framework/pull/1567**

Same branch, same two `dakeraMemory` files, now with `Signed-off-by`. Apologies for the noise, @Tomas2D — the review can continue on #1567.

Issue #1514, opened 2026-06-18
🚨 CRITICAL: Unsafe Function Deserialization RCE (CVSS 10.0) - GHSA-phhm-7927-g88p


jjasghar:
# 🚨 CRITICAL EMERGENCY: Unsafe Function Deserialization RCE (CVSS 10.0) - GHSA-phhm-7927-g88p

**SECURITY ADVISORY STATUS:** Triage (Not yet published - URGENT ACTION REQUIRED)  
**GHSA ID:** GHSA-phhm-7927-g88p  
**Severity:** CRITICAL (CVSS 10.0 - Maximum Severity)  
**Vulnerability Type:** Remote Code Execution via Function() Constructor

---

## ⚠️ IMMEDIATE ATTENTION REQUIRED

This is a **CVSS 10.0** remote code execution vulnerability that allows attackers to execute arbitrary JavaScript code in the Node.js process through crafted serialized payloads. The security advisory was reported on June 4, 2026, and is currently in "triage" state, meaning it has not yet been publicly disclosed.

**Time-sensitive action required before public disclosure!**

---

## Vulnerability Description

The TypeScript `Serializer.deserialize()` method in beeai-framework registers `Function` as a deserializable class without implementing any allowlist or validation mechanism. This creates a critical RCE vulnerability where:

1. An attacker crafts a JSON payload with a malicious `__value.fn` string
2. The serializer passes this string directly to the `Function()` constructor
3. The Function() constructor immediately executes the code as JavaScript
4. **Result:** Complete remote code execution in the Node.js process with no authentication required

### Attack Vector
```javascript
// Malicious payload example (simplified)
{
  "__type": "Function",
  "__value": {
    "fn": "require('child_process').execSync('rm -rf /')"
  }
}
```

When deserialized, this executes arbitrary code in the server context.

---

### Severity: CRITICAL (CVSS 10.0)

**Exploitation Requirements:**
- ✅ No authentication required
- ✅ User interaction not required  
- ✅ Remote attack vector
- ✅ Complete compromise of confidentiality, integrity, and availability

**Potential Consequences:**
- 🚨 Full system compromise via arbitrary code execution
- 🚨 Data exfiltration from production environments
- 🚨 Deployment of malware or backdoors
- 🚨 Lateral movement within infrastructure
- 🚨 Supply chain attacks through compromised framework usage
- 🚨 Complete service disruption

**Affected Versions:** beeai-framework@0.1.29 and likely other recent versions

---

[...]

PR #1551, opened 2026-07-04
fix(serializer): disable function deserialization by default (GHSA-phhm-7927-g88p)

SahebSandhuSingh:
## Summary

Fixes the CVSS 10.0 remote code execution vulnerability reported in #1514 (GHSA-phhm-7927-g88p). `Serializer.deserialize()` / `Serializable.fromSerialized()` previously reconstructed `Function` values by passing the saved source string directly into the `Function()` constructor, with no restriction on where that data came from. An attacker could craft a JSON payload that, once deserialized, executed arbitrary code with no authentication required.

## Root cause

`typescript/src/serializer/serializer.ts` registers `Function` as a deserializable type by default. `Serializable.fromSerialized()` — a public method inherited by every `Serializable` subclass (agents, memory, tools, caches, etc.) — accepts a raw string from anywhere and deserializes it through this global registry, so the vulnerability was reachable from any public API surface that calls `fromSerialized()` on untrusted input.

## Why not just remove `Function` from the registry entirely

I traced every call site of `deserialize`/`fromSerialized` and found function deserialization is a deliberate, actively-used feature — many `Tool`/`Agent`/`Memory` classes carry live functions (e.g. `DynamicTool`'s `handler`), and the framework's own test suite round-trips them via `serialize()`/`fromSerialized()`. A blanket removal would be a much larger breaking change than the vulnerability itself requires.

## Fix

Function *reconstruction* (`fromPlain`, the actually dangerous direction — it calls `Function()`) is now disabled by default, scoped via `AsyncLocalStorage` (consistent with the existing pattern in `context.ts`) so it doesn't leak across concurrent/unrelated calls:

- `Serializer.deserialize()` / `deserializeWithMeta()` accept a new `{ allowFunctionDeserialization?: boolean }` option, default `false`.
- `DeserializeOptions` (used by `Serializable.fromSerialized()`) gets the same option, threaded through.
- `Function`'s `toPlain` (describing a function — safe, never executes anything) is untouched.
- `src/middleware/trajectory.ts` explicitly opts in for its internal serialize→deserialize round-trip used for log formatting, since that data is self-generated and never attacker-controlled.
- The framework's own test suite (11 files) explicitly opts in wherever it legitimately round-trips functions in tests (tools, memory, agents), proving the feature still works when trust is deliberate.

## Testing

- Added a new regression test in `serializer.test.ts`: deserializing a function without opting in throws; with `{ allowFunctionDeserialization: true }` it still works correctly.
- All existing tests updated and passing (291/291, excluding e2e/example tests that require live external services).
- `tsc --noEmit` and `eslint` both pass clean.

## Scope note

This closes the specific vulnerability reported in #1514. It does not audit other registered types in the serializer for similar risks — that felt out of scope for this fix, but may be worth a separate pass.

Opening as a draft for review given the security sensitivity here — happy to adjust the design (e.g. naming, default behavior) based on feedback.

### DIFF
--- a/typescript/src/adapters/langchain/tools.test.ts
+++ b/typescript/src/adapters/langchain/tools.test.ts
@@ -34,7 +34,9 @@ describe("Langchain Tools", () => {
     });
 
     const serialized = await instance.serialize();
-    const deserialized = await LangChainTool.fromSerialized(serialized);
+    const deserialized = await LangChainTool.fromSerialized(serialized, {
+      allowFunctionDeserialization: true,
+    });
     verifyDeserialization(instance, deserialized, undefined, [], ["tool.schema"]);
   });
 });
--- a/typescript/src/agents/experimental/streamlit/agent.test.ts
+++ b/typescript/src/agents/experimental/streamlit/agent.test.ts
@@ -15,7 +15,9 @@ describe("Streamlit agent", () => {
       memory: new UnconstrainedMemory(),
     });
     const serialized = await instance.serialize();
-    const deserialized = await StreamlitAgent.fromSerialized(serialized);
+    const deserialized = await StreamlitAgent.fromSerialized(serialized, {
+      allowFunctionDeserialization: true,
+    });
     verifyDeserialization(instance, deserialized);
   });
 });
--- a/typescript/src/internals/serializable.ts
+++ b/typescript/src/internals/serializable.ts
@@ -19,6 +19,12 @@ interface SerializableStructure<T> {
 
 export interface DeserializeOptions {
   extraClasses?: SerializableClass<unknown>[];
+  /**
+   * Function deserialization is disabled by default because it can execute
+   * arbitrary code from the payload (see GHSA-phhm-7927-g88p). Only set this
+   * to `true` if you fully trust the source of the serialized data.
+   */
+  allowFunctionDeserialization?: boolean;
 }
 
 export abstract class Serializable<T = unknown> {
@@ -54,6 +60,8 @@ export abstract class Serializable<T = unknown> {
     const { __root } = await Serializer.deserializeWithMeta<SerializableStructure<T>>(
       value,
       options?.extraClasses,
+      false,
+      { allowFunctionDeserialization: options?.allowFunctionDeserialization },
     );
     if (!__root.target) {
       // eslint-disable-next-line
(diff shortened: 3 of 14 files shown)


Tomas2D:
Thanks @SahebSandhuSingh — this is exactly the right fix. Making function deserialization opt-in (secure by default), the clear `SerializerError`, and especially scoping the flag through `AsyncLocalStorage` so it can't leak across concurrent deserializations — all excellent, and the tests cover both the default-reject and opt-in paths.

CI is red on **Prettier only**: _"Code style issues found in 12 files."_ Run `yarn prettier --write .` from `typescript/` and commit — that should turn it green.

One note for the release notes: this is a **behavioral breaking change** — agents/tools/memory that serialize function fields will now throw on `fromSerialized()` unless the caller passes `{ allowFunctionDeserialization: true }`. That's the correct remediation for the RCE, but worth flagging as breaking (e.g. `fix!:` / a `BREAKING CHANGE:` footer) so it's documented. Once Prettier is happy I'll approve and merge.

SahebSandhuSingh:
@Tomas2D I have rectified the error kindly check

Tomas2D:
Thanks — Prettier's happy now, but CI is red on a **unit test**: `src/serializer/serializer.test.ts > Serializer > Handles various function definitions` fails with `AssertionError: expected Promise{…} to deeply equal Promise{…}` (1 failed / 290 passed).

That "Promise deeply equal Promise" shape usually means a `deserialize(...)` result isn't being awaited, so the assertion compares two unresolved Promises — likely a spot where the new `options` argument was threaded through but the call lost its `await` (or the inline-snapshot value needs regenerating). Could you run `yarn vitest run src/serializer/serializer.test.ts` locally, fix that case, and push? Everything else (design, opt-in gating, the new reject test) looks great — this is the last thing between us and merging the RCE fix.

SahebSandhuSingh:
@Tomas2D It's done, kindly check

Tomas2D (approved):
Verified: function deserialization is now opt-in (secure by default), the AsyncLocalStorage scoping is the right call, the unit test regression is fixed, and CI is fully green. This closes the CVSS 10.0 RCE (#1514) and its duplicates (#1562, #1563). Approving. Thanks @SahebSandhuSingh for seeing it through! 🐝 (Note for release: this is a behavioral breaking change — callers deserializing function-valued fields must now pass `{ allowFunctionDeserialization: true }`.)

PR #1559, opened 2026-07-07
fix(internals): compare both matrices' column counts in cosineSimilarityMatrix

### LINKED ISSUE #1558: cosineSimilarityMatrix column-count guard compares matrixA to itself (never fires)
### Summary

`cosineSimilarityMatrix` (`typescript/src/internals/helpers/math.ts`) has a column-count guard that can never fire, because it compares `matrixA[0]` to itself:

```ts
if ((matrixA[0]?.length ?? 0) !== (matrixA[0]?.length ?? 0)) {
  throw new ValueError("Matrices must have the same number of columns.");
}
```

The error message shows the intent is to compare A vs B, so the second operand should be `matrixB[0]`.

### Impact

Matrices with differing column counts bypass validation. For an empty `matrixB` the function returns `[[]]` instead of raising; for other mismatches it eventually throws the less-specific inner "Vectors must have equal length" error, and the intended guard message is unreachable.

### Fix

```ts
if ((matrixA[0]?.length ?? 0) !== (matrixB[0]?.length ?? 0)) {
```

PR to follow.

Osamaali313:
Closes #1558

## Problem

`cosineSimilarityMatrix` (`typescript/src/internals/helpers/math.ts`) validates that both matrices have the same column count, but the guard compares `matrixA[0]` to **itself**, so it can never fire:

```ts
if ((matrixA[0]?.length ?? 0) !== (matrixA[0]?.length ?? 0)) {
  throw new ValueError("Matrices must have the same number of columns.");
}
```

Mismatched column counts bypass validation — an empty `matrixB` returns `[[]]` instead of raising, and other mismatches fall through to the less-specific inner `"Vectors must have equal length"` error, leaving the intended guard message unreachable.

## Fix

Compare `matrixB[0]` (a one-token typo fix), as the error message intends:

```ts
if ((matrixA[0]?.length ?? 0) !== (matrixB[0]?.length ?? 0)) {
```

## Tests

Added `math.test.ts` asserting that mismatched column counts (`[[1,2]]` vs `[[1,2,3]]`, and vs `[]`) throw `"Matrices must have the same number of columns."`, and that a matching pair still computes similarity. Before the fix the mismatch cases don't raise that error; after, they do.


Tomas2D:
Good catch. Thank you!

Tomas2D (approved):

PR #1560, opened 2026-07-08
fix(chat): clear cached invalid tool call before retry

### LINKED ISSUE #1554: Cache entry is not cleared on ChatModelToolCallError
**Describe the bug**
When `ChatModelToolCallError` is encountered, the cache entry is not cleared, unlike when `EmptyChatModelResponseError` or general `Exception` is raised.

This means that the agent may get into a loop after a single failed tool call. 

**To Reproduce**

1. Create a mock ChatModel that produces one malformed tool call, followed by a correct one.
2. Attach `SlidingCache`, or any other cache, to it.
3. Invoke the model with number of retries > 1
4. See how the first, malformed tool call is cached and model is called only once.

**Expected behavior**
Cache entry with faulty tool call should be cleared, and model called again, to hopefully produce a correct response.

**Screenshots / Code snippets**
https://github.com/i-am-bee/beeai-framework/blob/main/python/beeai_framework/backend/chat.py#L552-L591

**Set-up:**

- Bee version: 0.1.81 and probably all up to and including main
- Model provider: vLLM (self hosted with gemma)

gyx09212214-prog:
## Summary

- Clear the chat model cache entry before retrying `ChatModelToolCallError` recovery.
- Add a regression test that covers a cached malformed tool call followed by a successful retry.

## Why this matters

When invalid tool-call recovery appends correction messages, the retry should make a fresh model call. If the malformed response remains cached, the retry reads the same bad response again and the model never sees the correction messages.

Fixes #1554.

## Validation

- `python -m pytest tests/backend/test_chatmodel_retry_cache.py -q -o addopts=`
- `python -m ruff format --check beeai_framework/backend/chat.py tests/backend/test_chatmodel_retry_cache.py`
- `python -m ruff check beeai_framework/backend/chat.py tests/backend/test_chatmodel_retry_cache.py`

### DIFF
--- a/python/beeai_framework/backend/chat.py
+++ b/python/beeai_framework/backend/chat.py
@@ -570,6 +570,7 @@ class ChatModel(Runnable[ChatModelOutput]):
                             {"tempMessage": True},
                         )
                     )
+                    await cache_entry.delete()
                 elif self.retry_on_empty_response and isinstance(e, EmptyChatModelResponseError):
                     model_input.messages = model_input.messages.copy()
                     model_input.messages.append(AssistantMessage("", {"tempMessage": True}))
--- a/python/tests/backend/test_chatmodel_retry_cache.py
+++ b/python/tests/backend/test_chatmodel_retry_cache.py
@@ -1,0 +1,47 @@
+# Copyright 2025 BeeAI a Series of LF Projects, LLC
+# SPDX-License-Identifier: Apache-2.0
+
+from collections.abc import AsyncGenerator
+
+import pytest
+
+from beeai_framework.backend.chat import ChatModel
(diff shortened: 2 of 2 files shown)


vivek41-glitch:
Git commit with sign-off, and dco will pass

Tomas2D:
Thanks @gyx09212214-prog — this carries the correct one-line fix for #1554. I'm closing it in favor of #1555, which makes the identical change but is fully green in CI (including the build/test suite) and has a test that reproduces the issue's exact malformed→corrected scenario. This PR's CI didn't run the test suite and DCO is failing. No wasted effort — the fix you both found is landing via #1555. Thanks for chasing it down!

PR #1561, opened 2026-07-08
fix(emitter): distinguish regex listeners by pattern OR flags in off()

Diwak4r:
## Summary
- `Emitter.off()` matched compiled-regex listeners with an `and` condition over pattern AND flags, so two listeners sharing a pattern but differing in flags (e.g. `re.IGNORECASE`) could not be removed independently — removing one removed the match test for both.
- Changed to `or` so a listener is only excluded when BOTH pattern and flags differ.

## Test plan
- Added `test_regex_off_distinguishes_flags` to `python/tests/test_emitter.py`.
- Fail-before/pass-after verified locally.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/python/beeai_framework/emitter/emitter.py
+++ b/python/beeai_framework/emitter/emitter.py
@@ -312,7 +312,7 @@ def _match_listener(
             return False
 
         if isinstance(matcher, re.Pattern) and isinstance(listener.raw, re.Pattern):
-            if matcher.pattern != listener.raw.pattern and matcher.flags != listener.raw.flags:
+            if matcher.pattern != listener.raw.pattern or matcher.flags != listener.raw.flags:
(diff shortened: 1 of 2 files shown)


Tomas2D:
Thanks @Diwak4r — this is a correct fix for the `off()` regex bug (the `and`→`or` on the pattern/flags comparison). I'm closing it as a duplicate of #1566, which makes the identical source change and carries a test for the common differing-pattern case. Your test for the *flags-differ* scenario is a nice complement — feel free to propose it on top of #1566, it'd be a welcome addition. Appreciate the fix!

PR #1571, opened 2026-07-24
docs(tools): add action-bound governance example

jw-ond:
## Summary

Adds a Python tools example showing how to bind a tool-side governance checkpoint to a canonical action hash before performing an external side effect.

The example is intentionally vendor-neutral and runs with an in-process checkpoint so it can be read and executed without any external service. A production deployment could replace the local checkpoint with an HTTP service that returns the same kind of decision object.

## What it demonstrates

- Building a canonical action envelope from the proposed tool action
- Hashing that envelope before execution
- Calling a checkpoint before the side effect is performed
- Failing closed if the checkpoint decision is not bound to the same action hash
- Returning structured outcomes for `approve`, `require_approval`, and `reject`

## Verification

```sh
PYTHONPATH=python /tmp/beeai-action-bound-venv/bin/python -m py_compile python/examples/tools/action_bound_governance.py
/tmp/beeai-action-bound-venv/bin/ruff check python/examples/tools/action_bound_governance.py python/examples/README.md
/tmp/beeai-action-bound-venv/bin/ruff format --check python/examples/tools/action_bound_governance.py
PYTHONPATH=python /tmp/beeai-action-bound-venv/bin/python python/examples/tools/action_bound_governance.py
```

The example prints one completed action, one approval-required action, and one rejected action.

### DIFF
--- a/python/examples/README.md
+++ b/python/examples/README.md
@@ -58,6 +58,7 @@ This repository contains examples demonstrating the usage of the BeeAI Framework
 ## Tools
 
 - [`decorator.py`](/python/examples/tools/decorator.py): Tool creation using decorator
+- [`action_bound_governance.py`](/python/examples/tools/action_bound_governance.py): Tool-side governance checkpoint bound to a canonical action hash
 - [`duckduckgo.py`](/python/examples/tools/duckduckgo.py): DDG Search Tool for searching the web
 - [`openmeteo.py`](/python/examples/tools/openmeteo.py): Open-Meteo Tool for retrieving weather data
 - [`custom/xquik.py`](/python/examples/tools/custom/xquik.py): Xquik API tool for searching X posts
--- a/python/examples/tools/action_bound_governance.py
+++ b/python/examples/tools/action_bound_governance.py
@@ -1,0 +1,177 @@
+import asyncio
+import hashlib
+import json
+import sys
+from typing import Literal, Protocol
+
+from pydantic import BaseModel, Field
+
+from beeai_framework.context import RunContext
+from beeai_framework.emitter import Emitter
+from beeai_framework.errors import FrameworkError
+from beeai_framework.tools import JSONToolOutput, Tool, ToolRunOptions
+
+Verdict = Literal["approve", "require_approval", "reject"]
+
+
+class GovernedCustomerUpdateInput(BaseModel):
+    case_id: str = Field(description="Support case identifier.")
+    update_text: str = Field(description="The proposed customer-support update.")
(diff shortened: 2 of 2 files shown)


Tomas2D (commented):
Thanks — the code itself is clean and idiomatic: correct `Tool` subclass (`name` / `description` / `input_schema` / `_create_emitter` / `_run`), canonical JSON before hashing, `Protocol` for the checkpoint so the in-process implementation is obviously swappable, and no new dependencies. It also runs standalone with no LLM, which is a nice property for an example.

My hesitation is about scope rather than quality, and I'd like to settle it before merging.

**Does this belong in `examples/tools/`?** Everything else there demonstrates a framework feature — `decorator.py` shows `@tool`, `openmeteo.py` shows a real tool. This demonstrates an *architectural pattern* built on top of the framework, and 177 lines of it. A reader looking for "how do I write a tool" gets a support-ticket domain model, a verdict enum, two hash helpers and a checkpoint protocol before reaching anything framework-specific.

**The hash binding doesn't do what the framing implies.** `LocalActionCheckpoint` receives `action_hash` and echoes it straight back, so `decision.action_hash != action_hash` can never be true and that branch is unreachable in the example as written. Against an in-process checkpoint the binding is decorative — it only earns its keep when the decision crosses a trust boundary and comes back signed. Since the docstring points at "an HTTP call to an external governance service" as the real deployment, the example would be more honest either demonstrating that (a checkpoint that returns a mismatched hash, showing the guard firing) or dropping the hash and being a plainer pre-execution approval example.

Two smaller things if it does land:
- The framework already has approval primitives — `RequirementAgent` with `AskPermissionRequirement` (`agents/requirement/requirements/ask_permission.py`), and now the middleware route in #1610. The example should say why a tool-side checkpoint is the right layer rather than those, otherwise it reads as a competing pattern.
- `sha256_text(canonical_json(...))` over a `decision_id` derived from the verdict and reason is a content hash, not a decision identifier — a second identical decision collides with the first. Fine for a demo, but worth a comment so nobody copies it into an audit log.

Could you say a bit about the motivating use case? If there's a real deployment behind this, that would help me judge whether it's an example or a docs page.

jw-ond:
Rebased this branch onto current main to clear the stale-base CI failure. The previous failing job stopped during Poetry build with a TOML parsing error in python/pyproject.toml, outside this PR's two-file docs/example diff. On the rebased branch, python/pyproject.toml parses cleanly with Python 3.11 tomllib; current visible checks on the new head are DCO and labeler, both passing.

PR #1576, opened 2026-07-29
feat(adapters): add MiniMax regional (CN) endpoint selection

octo-patch:
Reason: The MiniMax adapters only defined the global endpoint, so CN users had to discover and type the regional base URL by hand; this adds a first-class global/CN region selection path.

## What changed

MiniMax serves the same OpenAI-compatible API from two regional gateways:

- global: `https://api.minimax.io/v1` (unchanged default)
- CN: `https://api.minimaxi.com/v1`

Both the Python and TypeScript adapters previously hard-coded only the global base URL. This PR adds an explicit region selection path in both languages while preserving the existing default behavior.

### Python (`beeai_framework/adapters/minimax`)

- Add `MINIMAX_API_BASE_CN` and a `resolve_minimax_base_url()` helper backed by a `MINIMAX_REGION_BASE_URLS` map (`global`/`global_en` and `cn`/`cn_zh`).
- `MiniMaxChatModel` accepts a `region` argument, falling back to the `MINIMAX_API_REGION` environment variable.
- Base URL precedence is explicit and unchanged for existing users: `base_url` argument -> `MINIMAX_API_BASE` env -> region endpoint -> global default.
- Export the new constants and helper from the package.

### TypeScript (`typescript/src/adapters/minimax`)

- Export `MINIMAX_API_BASE`, `MINIMAX_API_BASE_CN`, `MINIMAX_REGION_BASE_URLS`, the `MiniMaxRegion` type, and `resolveMiniMaxBaseURL()`.
- `MiniMaxClientSettings` gains an optional `region`; the client resolves the base URL with the same precedence and strips `region` before forwarding settings to the underlying OpenAI-compatible provider.

### Docs & examples

- Document `MINIMAX_API_REGION` and the region option in the backend module docs.
- Show the region option in the Python and TypeScript provider examples.

### Tests

- Add parity unit tests in both languages covering region resolution (global/CN aliases, case-insensitivity, empty default, unknown-region error) and the base-URL precedence rules.

## Checks

- Python: `ruff check` (clean), `ruff format --check` (clean), `pytest tests/adapters/minimax/test_minimax_chat.py` -> 27 passed.
- TypeScript: `tsc --noEmit` (clean), `vitest run tests/e2e/adapters/minimax.test.ts` -> 16 passed, `prettier --check` (clean), `eslint` (clean).

### DIFF
--- a/docs/src/content/docs/modules/backend.mdx
+++ b/docs/src/content/docs/modules/backend.mdx
@@ -37,7 +37,7 @@ The following table depicts supported providers. Each provider requires specific
 | Google Gemini  |  ✅  | ✅        | GEMINI_CHAT_MODEL<br/>**GEMINI_API_KEY**<br/>GEMINI_API_HEADERS |
 | MistralAI      |  ✅  | ✅        | MISTRALAI_CHAT_MODEL<br/>MISTRALAI_EMBEDDING_MODEL<br />**MISTRALAI_API_KEY**<br />MISTRALAI_API_BASE |
 | Transformers   |  ✅  | ✅        | TRANSFORMERS_CHAT_MODEL<br/>HF_TOKEN|
-| MiniMax        |  ✅  | ❌        | MINIMAX_CHAT_MODEL<br/>**MINIMAX_API_KEY**<br/>MINIMAX_API_BASE<br/>MINIMAX_API_HEADERS |
+| MiniMax        |  ✅  | ❌        | MINIMAX_CHAT_MODEL<br/>**MINIMAX_API_KEY**<br/>MINIMAX_API_BASE<br/>MINIMAX_API_REGION<br/>MINIMAX_API_HEADERS |
 
 <Tip>
 If you don't see your provider raise an issue [here](https://github.com/i-am-bee/beeai-framework/issues).
@@ -47,6 +47,10 @@ or [TypeScript](https://github.com/i-am-bee/beeai-framework/blob/main/typescript
 
 <Note>
 	Google Gemini, MistralAI, and Transformers are supported in Python only. The Transformers chat model does not support tool calling.
+</Note>
+
+<Note>
+	MiniMax serves the same API from two regional gateways. Select one with `MINIMAX_API_REGION` (or the `region` option): `global` (default, `https://api.minimax.io/v1`) or `cn` (`https://api.minimaxi.com/v1`). An explicit `MINIMAX_API_BASE` / `base_url` always takes precedence, so you no longer have to look up and type the regional URL by hand.
 </Note>
 
 ---
--- a/python/beeai_framework/adapters/minimax/__init__.py
+++ b/python/beeai_framework/adapters/minimax/__init__.py
@@ -1,6 +1,16 @@
 # Copyright 2025 © BeeAI a Series of LF Projects, LLC
 # SPDX-License-Identifier: Apache-2.0
 
-from beeai_framework.adapters.minimax.backend.chat import MiniMaxChatModel
+from beeai_framework.adapters.minimax.backend.chat import (
+    MINIMAX_API_BASE,
+    MINIMAX_API_BASE_CN,
+    MiniMaxChatModel,
+    resolve_minimax_base_url,
+)
 
-__all__ = ["MiniMaxChatModel"]
+__all__ = [
+    "MINIMAX_API_BASE",
+    "MINIMAX_API_BASE_CN",
+    "MiniMaxChatModel",
+    "resolve_minimax_base_url",
+]
--- a/python/beeai_framework/adapters/minimax/__init__.py
+++ b/python/beeai_framework/adapters/minimax/__init__.py
@@ -1,16 +1,6 @@
 # Copyright 2025 © BeeAI a Series of LF Projects, LLC
 # SPDX-License-Identifier: Apache-2.0
 
-from beeai_framework.adapters.minimax.backend.chat import (
-    MINIMAX_API_BASE,
-    MINIMAX_API_BASE_CN,
-    MiniMaxChatModel,
-    resolve_minimax_base_url,
-)
+from beeai_framework.adapters.minimax.backend.chat import MiniMaxChatModel
 
-__all__ = [
-    "MINIMAX_API_BASE",
-    "MINIMAX_API_BASE_CN",
-    "MiniMaxChatModel",
-    "resolve_minimax_base_url",
-]
+__all__ = ["MiniMaxChatModel"]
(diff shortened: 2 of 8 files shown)


Tomas2D:
Thanks for adding MiniMax regional (CN) endpoint selection — appreciated.

**Good news: the CI failure here isn't your code.** Your changes passed lint, formatting, and type-checks cleanly. The only failing step is `poetry build`, which is dying on:

```
Invalid TOML file python/pyproject.toml: Key "lint" already exists.
```

That was a pre-existing TOML block-ordering bug in `pyproject.toml` on the base branch you forked from (`[tool.pyrefly]` was sitting between the `[tool.ruff.lint.*]` tables). **It's already fixed on `main`.**

To get a green build, please just **rebase / merge the latest `main`** into your branch:

```bash
git fetch upstream    # or 'origin' if you branched from the main repo
git rebase upstream/main
git push --force-with-lease
```

That should be all it takes — CI should go green afterward. Please don't be discouraged by the red ❌; your actual contribution is in good shape. Thanks for your patience, and looking forward to getting this merged! 🐝

octo-patch:
Updated the branch with the latest `main` to pick up the Python build configuration fix. I ran `poetry build`, the MiniMax Python tests (27 passed), the MiniMax TypeScript tests (16 passed), and `tsc --noEmit`.

Tomas2D (commented):
Thanks. The shape of this is right — region-name → base URL, with an explicit `base_url` / `MINIMAX_API_BASE` still winning — and the test coverage on both sides is good. Three things before it can go in.

**1. Please cite the source for the two URLs (blocking).**

```python
MINIMAX_API_BASE    = "https://api.minimax.io/v1"
MINIMAX_API_BASE_CN = "https://api.minimaxi.com/v1"
```

This is the highest-risk line in the PR and I can't verify it from the diff. If the mapping is inverted, we silently route traffic to the wrong jurisdiction, which is exactly the failure this feature exists to prevent — and the tests only assert the constants against themselves, so they'd stay green either way. Please link the MiniMax documentation page that states which gateway serves which region. Note `api.minimax.chat` has also been in circulation historically, so "the other one is CN" isn't safe to infer.

**2. Python and TypeScript disagree on when an invalid region raises.**

In Python the fallback is an eagerly-evaluated argument:

```python
fallback=resolve_minimax_base_url(region),
```

so `MiniMaxChatModel(base_url="https://proxy.example.com/v1", region="mars")` raises `ValueError` even though `region` is never used. TypeScript short-circuits instead (`settings.baseURL || getEnv(...) || resolveMiniMaxBaseURL(...)`), so the same call succeeds there. Pick one behaviour and match it in both. I'd resolve the region lazily in Python — only when no `base_url` and no env var were supplied — and add the `base_url` + invalid-region case as a test, since the current `test_explicit_base_url_overrides_region` passes a *valid* region and so doesn't cover this.

**3. Drop the `global_en` / `cn_zh` aliases.**

Four accepted spellings for two endpoints is API surface we have to keep forever. Unless these are identifiers MiniMax itself publishes (in which case, same request as #1: link them), `global` and `cn` are enough.

Minor: `resolve_minimax_base_url` and the two constants are now exported from `beeai_framework.adapters.minimax`. Is the resolver meant to be public API? No other adapter exports a URL helper, and once it's in `__all__` we own it.

One process note — this PR has no linked issue and adds a user-facing env var (`MINIMAX_API_REGION`) across both languages. Could you open one, or describe the concrete setup that needs this?

octo-patch:
Thanks for the review. I made the Python region lookup lazy behind explicit or environment base URLs, removed the `global_en` and `cn_zh` aliases, kept the Python resolver and constants internal, and pushed parity tests. MiniMax official CLI documentation identifies `api.minimax.io` as Global and `api.minimaxi.com` as CN ([source](https://github.com/MiniMax-AI/cli/blob/1e9b58167d302757a1a88347f51d031369134178/README.md#L20-L26)), and its endpoint tests show the `/v1` paths for both hosts ([source](https://github.com/MiniMax-AI/cli/blob/1e9b58167d302757a1a88347f51d031369134178/test/client/endpoints.test.ts#L4-L12)); the concrete use case is one application configuration deployed against separate Global and China MiniMax accounts. I ran the MiniMax Python tests (29 passed), MiniMax TypeScript tests (19 passed), Ruff, ESLint, Prettier, and `tsc --noEmit`.

Tomas2D (commented):
This addresses everything, thank you — and the citation is exactly what I needed. The MiniMax CLI README confirming `api.minimax.io` = Global and `api.minimaxi.com` = CN settles the blocking question, and it's now recorded in the thread for whoever touches this next.

Verified in the diff:
- Fallback is lazy — `fallback=None if has_base_url else resolve_minimax_base_url(region)`, so an invalid region no longer raises when `base_url` or the env var wins. Python and TypeScript now agree.
- `global_en` / `cn_zh` are gone, and you added tests asserting they're *rejected*, which is better than just deleting them.
- The resolver and constants stayed internal — `adapters/minimax/__init__.py` is out of the diff entirely.

**One thing left, and it's not code.** The TypeScript job fails at `typescript:check:commitlint`:

```
⧗   input: chore: merge latest main for CI
✖   message must be signed off [signed-off-by]
```

Commit `14e2f5c4` has no `Signed-off-by` trailer. The Python job passes and DCO passes; it's only commitlint. Easiest fix is to rebase and drop that merge commit (the branch is already current), or amend it with a sign-off:

```
git rebase --signoff origin/main
git push --force-with-lease
```

Once that's green I'll merge it.

Tomas2D (commented):
Reviewed the code — the approach is good and the Python/TypeScript sides line up cleanly. Precedence is consistent in both: explicit `base_url`/`baseURL` → `MINIMAX_API_BASE` → region → global default. Destructuring `region` out of the settings in `MiniMaxClient.create()` so it isn't forwarded to `createOpenAI` is the right touch, and the test coverage on both sides is thorough.

**Blocker (mechanical, not code):** CI is red on commitlint, not on your changes:

```
⧗   input: chore: merge latest main for CI
✖   message must be signed off [signed-off-by]
```

Commit `14e2f5c` ("chore: merge latest main for CI") has no `Signed-off-by` trailer. Rebasing onto current `main` to drop that merge commit is the simplest fix; `git rebase --signoff` works too if you'd rather keep it. The Python job and the docs build both pass, so that one check is all that stands between this and a merge.

**One behavioral note, worth a line in the docs at most:** an invalid `region` only raises when no base URL is set from any source — `MiniMaxChatModel(base_url=..., region="mars")` silently ignores the bad region (your `test_explicit_base_url_ignores_invalid_region` pins this). That's a defensible choice and it's symmetric across both languages, so I'd leave the behavior alone; it just isn't obvious from the docstring, which reads as though `region` is always validated.

PR #1582, opened 2026-08-03
fix(emitter): keep an unset group_id unset when cloning

LHMQ878:
### Which issue(s) does this pull-request address?

Closes: #1581

### Description

`Emitter.clone()` built the clone with the group id passed positionally through `str()`:

```python
cloned = Emitter(
    str(self._group_id),
    ...
)
```

`group_id` is `str | None` and defaults to `None`, so when it was unset the clone received the literal string `"None"`. `_create_event()` stamps `group_id=self._group_id` onto every `EventMeta`, and `child()` does `group_id=group_id or self._group_id`, so the synthetic value propagated to every event emitted by the clone and by every emitter descended from it.

Measured on `main` (`21284d7`, Python 3.12.10):

| expression | before | after |
|---|---|---|
| `Emitter(namespace=["app"])._group_id` | `None` | `None` |
| `(await that.clone())._group_id` | `'None'` | `None` |
| `cl._group_id is None` | `False` | `True` |
| `bool(cl._group_id)` | `True` | `False` |
| `event.group_id` from the clone | `'None'` | `None` |
| `(await that.clone()).child(...)._group_id` | `'None'` | `None` |
| clone with `group_id="run-42"` | `'run-42'` | `'run-42'` |

Agents are the common case rather than an edge case. All nine `clone()` implementations do `cloned.emitter = await self.emitter.clone()` — the four adapter agents (`a2a`, `acp`, `agentstack`, `watsonx_orchestrate`), the four core agents (`lite`, `react`, `requirement`, `tool_calling`), and `context.py:281` — and agents build their emitter via `Emitter.root().child(namespace=["agent", "tool_calling"], creator=self, events=...)` with no group id. Events off a cloned agent looked like this:

```
run.agent.tool_calling.start     group_id='None'
agent.tool_calling.done          group_id='None'
```

Anything grouping or filtering by `group_id` therefore saw a spurious `"None"` group for cloned agents: `if meta.group_id:` became true, `meta.group_id is None` became false, and a dict keyed by group id gained a `"None"` bucket. Forwarding `meta.group_id` to a tracing backend recorded the string as-is.

The fix passes the constructor arguments by keyword and drops the coercion, so the absent value stays absent. `creator` no longer needs the `if self.creator else None` guard — the attribute is already `object | None`, so the conditional was a no-op.

[...]

PR #1583, opened 2026-08-04
fix(emitter): preserve unset group id when cloning

### LINKED ISSUE #1581: Emitter.clone() turns an unset group_id into the literal string "None" (Python)
## Describe the bug

`Emitter.clone()` (Python) passes the group id positionally through `str()`:

```python
# python/beeai_framework/emitter/emitter.py
async def clone(self) -> "Emitter":
    cloned = Emitter(
        str(self._group_id),
        ...
    )
```

`group_id` is `str | None` and defaults to `None`. When it is unset, `str(None)` produces the literal string `"None"`, so the clone ends up with a truthy, meaningless group id instead of no group id at all.

Measured on `main` (`21284d7`, Python 3.12.10):

| expression | before | expected |
|---|---|---|
| `Emitter(namespace=["app"])._group_id` | `None` | `None` |
| `(await that.clone())._group_id` | `'None'` | `None` |
| `cl._group_id is None` | `False` | `True` |
| `bool(cl._group_id)` | `True` | `False` |
| `(await that.clone()).child(...)._group_id` | `'None'` | `None` |
| clone with `group_id="run-42"` | `'run-42'` | `'run-42'` |

The corruption is not confined to the clone object. `_create_event()` stamps `group_id=self._group_id` onto every `EventMeta`, and `child()` does `group_id=group_id or self._group_id`, so the fake `"None"` propagates to every event emitted by the clone and by every emitter descended from it:

```
run.agent.tool_calling.start     group_id='None'
agent.tool_calling.done          group_id='None'
```

## How to reproduce

```python
import asyncio
from beeai_framework.emitter.emitter import Emitter

async def main() -> None:
    emitter = Emitter(namespace=["app"])          # no group_id — the agent shape
    clone = await emitter.clone()
    print(repr(emitter._group_id), repr(clone._group_id))   # None 'None'

    clone.on("*", lambda _, event: print("event group_id:", repr(event.group_id)))
    await clone.emit("a", 1)                      # event group_id: 'None'
    print(repr(clone.child(namespace=["c"])._group_id))     # 'None'

asyncio.run(main())
```

## Impact

Every agent clone is affected. All nine `clone()` implementations do `cloned.emitter = await self.emitter.clone()`:

- `adapters/a2a/agents/agent.py`, `adapters/acp/agents/agent.py`, `adapters/agentstack/agents/agent.py`, `adapters/watsonx_orchestrate/agents/agent.py`
- `agents/lite/agent.py`, `agents/react/agent.py`, `agents/requirement/agent.py`, `agents/tool_calling/agent.py`
- `context.py:281`

and agents build their emitter with `Emitter.root().child(namespace=["agent", "tool_calling"], creator=self, events=...)` — no `group_id` — so the unset path is the normal path, not an edge case.

Consumers that group or filter events by `group_id` see a synthetic `"None"` group for cloned agents: `if meta.group_id:` becomes true, `meta.group_id is None` becomes false, and any dictionary keyed by group id gains a spurious `"None"` bucket. Anything forwarding `meta.group_id` to a tracing/observability backend records the string too.

## Why the existing test does not catch it

`python/tests/test_emitter.py::test_clone` constructs the emitter with `group_id="test_group"`. A real group id round-trips through `str()` unchanged, so the only broken path — the default `None` — is never exercised.

## TypeScript parity

The TypeScript emitter handles this correctly, so this is a Python-only divergence:

```ts
// typescript/src/emitter/emitter.ts
this.groupId = input?.groupId;                       // :60
groupId: input?.groupId ?? this.groupId,             // :80 (child)
groupId: this.groupId,                               // :254 (createSnapshot)
```

## Suggested fix

Pass the constructor arguments by keyword and drop the `str()` coercion, so `None` stays `None`. Happy to open a PR.

## Set-up

- `main` at `21284d7`
- Python 3.12.10

Diwak4r:
## Summary

`Emitter.clone()` built the cloned emitter by passing `str(self._group_id)`. Since `group_id` is `str | None` and defaults to `None`, an unset group id became the literal string `"None"` — a truthy, meaningless value that changed event semantics after cloning (e.g. `EventMeta.group_id` would suddenly be `"None"` instead of `None`).

The fix passes the group id through unchanged, so a clone of an emitter without a group id keeps `None`.

## Changes

- `python/beeai_framework/emitter/emitter.py`: `clone()` now passes `self._group_id` directly instead of `str(self._group_id)`.
- `python/tests/test_emitter.py`: regression tests — cloning preserves both an unset (`None`) and a set group id.

## Tests

`python -m pytest tests/test_emitter.py` → 18 passed. Dependent clone consumers (lite agent, token/summarize/sliding memory) → 33 passed. `ruff check` on changed files → clean.

Fixes #1581

### DIFF
--- a/python/beeai_framework/emitter/emitter.py
+++ b/python/beeai_framework/emitter/emitter.py
@@ -281,7 +281,7 @@ class Emitter:
 
     async def clone(self) -> "Emitter":
         cloned = Emitter(
-            str(self._group_id),
+            self._group_id,
             self.namespace.copy(),
             self.creator if self.creator else None,
             self.context.copy(),
--- a/python/tests/test_emitter.py
+++ b/python/tests/test_emitter.py
@@ -64,6 +64,26 @@ async def test_clone() -> None:
     assert clone.events is not emitter.events
 
 
+@pytest.mark.unit
+@pytest.mark.asyncio
+async def test_clone_preserves_unset_group_id() -> None:
+    emitter = Emitter(namespace=["namespace"])
+    assert emitter._group_id is None
+
+    clone = await emitter.clone()
+
+    assert clone._group_id is None
+
+
+@pytest.mark.unit
+@pytest.mark.asyncio
+async def test_clone_preserves_set_group_id() -> None:
+    emitter = Emitter(group_id="test_group", namespace=["namespace"])
+    clone = await emitter.clone()
+
+    assert clone._group_id == "test_group"
+
+
 class TestEventsPropagation:
     @pytest.mark.unit
     @pytest.mark.asyncio


Tomas2D:
Thanks @Diwak4r — this is a correct, minimal fix and your tests are good. 🙏

For full transparency: there are three PRs for #1581 (this, #1582, #1598). We're leaning toward #1582 because it converts the whole `clone()` call to keyword arguments, which additionally hardens it against future constructor-arg reordering — whereas this PR keeps the positional call that made the bug possible in the first place. Your fix is otherwise perfectly valid.

Please don't be discouraged — this was a real bug and your solution was right. If you'd like another target, the `good first issue` / `help wanted` labels have open items. Thank you for contributing! 🐝

Tomas2D:
#1581 has now been fixed and merged via #1582, so I'm closing this as resolved. Thank you again @Diwak4r — your fix was correct, it simply came down to choosing one of three valid PRs. Please do pick up another `good first issue` / `help wanted` item; we'd love to have you back. 🐝

PR #1575, opened 2026-07-29
chore(utils): Set units to seconds in the abort message

jpodivin:
### Which issue(s) does this pull-request address?

https://github.com/i-am-bee/beeai-framework/issues/1574

Closes: #

### Description

Simple change in reported time units.

#### General
- [x] I have read the appropriate contributor guide: [Python](https://github.com/i-am-bee/beeai-framework/blob/main/CONTRIBUTING.md)
/ [TypeScript](https://github.com/i-am-bee/beeai-framework/blob/main/CONTRIBUTING.md)
- [x] I have signed off on my commit: [Python instructions](https://github.com/i-am-bee/beeai-framework/blob/main/python/CONTRIBUTING.md#developer-certificate-of-origin-dco) / [TypeScript instructions](https://github.com/i-am-bee/beeai-framework/blob/main/typescript/CONTRIBUTING.md#developer-certificate-of-origin-dco)
- [x] Commit messages and PR title follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary)
- [x] Appropriate label(s) added to PR: `Python` for Python changes, `TypeScript` for TypeScript changes

#### Code quality checks
- [x] Code quality checks pass: `mise check` (`mise fix` to auto-fix)

#### Testing
- [x] Unit tests pass: `mise test:unit`
- [x] E2E tests pass: `mise test:e2e`
- [x] Tests are included (for bug fixes or new features)

#### Documentation
- [x] Documentation is updated
- [ ] Embedme embeds code examples in docs. To update after edits, run: Python `mise docs:fix`

PR #1584, opened 2026-08-04
fix(utils): report timeout duration in seconds

### LINKED ISSUE #1574: Timeout message states incorrect units
**Describe the bug**
Message for signal triggered by timeout uses milliseconds, when in fact the `call_later` method of the `AbstractEventLoop` uses seconds.

https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.call_later 

**To Reproduce**

1. Trigger timeout with `AbortSignal`
2. Observe the message

**Expected behavior**
Message should use correct units.

**Screenshots / Code snippets**

https://github.com/i-am-bee/beeai-framework/blob/21284d7f53d5a50e546350f371c69747bd6a176b/python/beeai_framework/utils/cancellation.py#L49

**Set-up:**

- Bee version: main

Diwak4r:
## What changed

`AbortSignal.timeout()` schedules its abort via `loop.call_later(duration, ...)`, where `duration` is expressed in **seconds** — but the abort reason message said `"Operation timed out after {duration} ms"`, which reports the wrong unit.

This PR fixes the message to use the correct unit (`s`), so e.g. `AbortSignal.timeout(0.01)` produces `"Operation timed out after 0.01 s"`.

## Why

The mismatch is user-facing: any code that surfaces the abort reason (for example, retry logic or error reporting) would display a misleading duration. The message now matches what `call_later` actually received.

## Tests

Added `python/tests/utils/test_cancellation.py`:

- `test_timeout_reason_uses_seconds` — asserts the reason for a sub-second timeout is reported in seconds, not milliseconds (fails on the old code, passes now).
- `test_manual_abort_reason` — guards the manual abort path.

Verified: new tests red → green; full `tests/utils` suite passes (53 passed); `ruff check` clean.

Note: one pre-existing failure in `tests/backend/test_chatmodel.py::test_cache_cleared_on_tool_call_error` reproduces on a clean `main` and is unrelated to this change.

Fixes #1574

### DIFF
--- a/python/beeai_framework/utils/cancellation.py
+++ b/python/beeai_framework/utils/cancellation.py
@@ -46,7 +46,7 @@ class AbortSignal(BaseModel):
         signal = cls()
 
         loop = asyncio.get_event_loop()
-        loop.call_later(duration, lambda *args: signal._abort(f"Operation timed out after {duration} ms"))
+        loop.call_later(duration, lambda *args: signal._abort(f"Operation timed out after {duration} s"))
 
         return signal
 
--- a/python/tests/utils/test_cancellation.py
+++ b/python/tests/utils/test_cancellation.py
@@ -1,0 +1,30 @@
+# Copyright 2025 © BeeAI a Series of LF Projects, LLC
+# SPDX-License-Identifier: Apache-2.0
+
+import asyncio
+
+import pytest
+
+from beeai_framework.utils import AbortSignal
+
+
+@pytest.mark.unit
+@pytest.mark.asyncio
+async def test_timeout_reason_uses_seconds() -> None:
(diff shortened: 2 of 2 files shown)


Tomas2D:
Thank you for this fix, @Diwak4r! 🙏 The same underlying issue (#1574 — timeout message reporting the wrong units) was just resolved and merged via #1575, so I'm closing this one as a duplicate.

Please don't let this discourage you — it was a real bug and your instinct to fix it was spot-on; #1575 simply landed first. We'd genuinely welcome more contributions from you. If you'd like another good starting point, take a look at the open issues labeled `good first issue` / `help wanted`. Thanks again for pitching in! 🐝

PR #1610, opened 2026-08-23
feat: add HOL Guard tool middleware

kantorcodes:
## Summary

Adds an optional `HOLGuardMiddleware` for BeeAI Python tool runs.

- evaluates tool name and validated input with the local `hol-guard` CLI before the BeeAI tool handler executes
- short-circuits the BeeAI run with a model-visible block result when Guard returns deny/review or evaluation fails
- fails closed on Guard startup/timeout by default, with explicit `fail_closed=False` opt-out
- supports nested tool runs through BeeAI's existing middleware/event propagation model
- adds unit coverage proving an allowed tool executes and a denied tool executes zero times
- adds a minimal runnable example

No BeeAI package dependency is added: users who want this middleware install HOL Guard separately (`pipx install hol-guard` or equivalent). Local Guard evaluation does not require Guard Cloud or sign-in.

## Why this fits BeeAI

BeeAI middleware already provides the pre-execution interception point needed for safety policy. `RunContext` middleware can stop a run before the target implementation is invoked, so this integration uses the existing framework boundary rather than adding a parallel policy abstraction.

HOL Guard's generic hook path accepts arbitrary harness identifiers and returns a non-zero status for actions that must not proceed. The middleware sends a `PreToolUse` payload under the `beeai` harness and only permits the underlying BeeAI `Tool._run` path on an allow result.

## Example

```python
from beeai_framework.middleware.hol_guard import HOLGuardMiddleware

result = await tool.run({"command": "echo hello"}).middleware(HOLGuardMiddleware())
```

With `match_nested=True` (the default), the same middleware can be applied at a parent run boundary and evaluate nested BeeAI tools.

## Validation

- added allow-path unit test
- added deny-path unit test asserting zero downstream executions
- added fail-closed / explicit fail-open tests
- added policy-action parsing and timeout validation tests
- syntax-checked the three added Python files with `python -m py_compile`

The commit includes the project-required DCO sign-off.


Tomas2D (commented):
Thanks for the contribution! I approved the CI run — `Lint & Build & Test` is now red on one concrete issue:

```
ERROR Object of class `object` has no attribute `run_id` [missing-attribute]
  --> beeai_framework/middleware/hol_guard.py:74:77
74 |         run_id = meta.trace.run_id if meta.trace and meta.trace.run_id else meta.creator.run_id
```

`EventMeta.creator` is typed as `object`, so the fallback branch doesn't type-check. `handle_start` already narrows it with `isinstance(meta.creator, RunContext)`, but that narrowing doesn't carry into `_build_payload`. Simplest fix is to pass the already-narrowed context (or the run id) in as a parameter rather than re-deriving it from `meta`.

Two design points I'd like to settle before merging:

1. **Placement.** `beeai_framework/middleware/` currently holds framework-generic middleware (`trajectory`, `stream_tool_call`). This one shells out to an external, vendor-specific `hol-guard` binary. Third-party integrations live under `beeai_framework/adapters/` in this repo — I think that's the right home for it. It also raises the maintenance question of who keeps the payload schema in sync with the HOL Guard CLI.

2. **`fail_closed=True` default.** If the binary isn't installed, every tool call is blocked with "HOL Guard could not start". That's the correct default for a guard, but it's worth calling out prominently in the docstring/example so it isn't diagnosed as a framework bug.

The rest reads well — the `bind()` cleanup pattern matches `trajectory.py`, and blocking via `start_event.output` correctly short-circuits execution in `RunContext` (`context.py:217`).

kantorcodes:
pushed those changes. can you approve the workflow runs when you get a chance?

Tomas2D (changes requested):
Thanks for this — the fail-closed default and the payload shape are both sensible. A few things before it can go in.

**Scope / maintainer call.** Everything currently under `python/beeai_framework/adapters/` is an LLM provider or an agent protocol (a2a, acp, mcp, …). This would be the first third-party *product* adapter vendored into the framework, and it hard-codes a `hol-guard` executable name and its CLI contract (`hook --harness beeai --json`, exit code 0 = allow). That's a maintainer decision, not just a code-review one — it may belong in a separate package rather than in-tree. Flagging so a maintainer weighs in.

**Bugs:**

1. `bind()` is not safe for concurrent runs. It starts with `while self._cleanups: self._cleanups.pop()()`, so binding to a second run unregisters the handler from the first. A middleware instance appended to `tool.middlewares` (as your own tests do) is shared across every run of that tool, so two concurrent runs leave one of them unguarded — the fail-open case this middleware exists to prevent. The cleanup list needs to be keyed per `RunContext`, or the handler registration needs to not tear down prior bindings.

2. Any non-zero exit is treated as a policy denial. If `hol-guard` is on PATH but crashes (bad args, exit 127, segfault), `_evaluate` returns `False` with a reason derived from stderr, and `fail_closed` is never consulted. The direction is safe, but it conflates "policy said no" with "the guard is broken", so operators running with `fail_closed=False` still get hard blocks from a broken binary. Consider having the guard report its decision in the JSON payload and reserving unexpected exit codes for `_availability_result`.

3. `test_hol_guard_fails_closed_when_unavailable_by_default`, `..._can_fail_open_...`, `..._formats_json_policy_action` and `..._requires_positive_timeout` are missing `@pytest.mark.unit`. CI runs `poetry run pytest -m unit` ([python/tasks.toml:96](https://github.com/i-am-bee/beeai-framework/blob/main/python/tasks.toml#L96)), so those four never execute.

**Smaller:** the three tests that do run only exercise `StubHOLGuardMiddleware`, which overrides `_evaluate` — the subprocess path, the timeout branch and the `OSError` branch have no coverage at all. And `python/examples/middleware/hol_guard.py` is missing the SPDX/copyright header the other examples carry.

Tomas2D:
Correction to my review above: scratch the SPDX header point —  files don't carry one, so your example matches the convention. The three bugs and the scope question stand.

Worth adding: that directory already has `prompt_injection.py`, `content_filter.py` and `secrets_detection.py` as guard-style middleware *examples*. That may be the natural home for this — an example showing how to wire an external policy process in, rather than a vendored adapter.

Tomas2D (approved):
This addresses everything from my earlier review — thanks for the rework. Clearing my changes-requested.

Going through the points in order:

**Scope** — relocating from `beeai_framework/adapters/` to `python/examples/middleware/` resolves it. It now sits alongside `prompt_injection.py`, `content_filter.py` and `secrets_detection.py` as an example of wiring a policy check into the tool-start boundary, rather than a vendored adapter the framework owns and has to maintain.

**Concurrency** — fixed properly. Keying on `ctx.run_id` in `_bound_run_ids` and unregistering from a `finish` handler is the right shape, and it's robust in a way that's worth calling out: `finish` is emitted from a `finally` block in [context.py:266-269](https://github.com/i-am-bee/beeai-framework/blob/main/python/beeai_framework/context.py#L266-L269), so the cleanup runs on the error and abort paths too, not just clean completion. No leak, and concurrent runs of the same tool no longer unbind each other.

**Test markers / stub-only coverage** — moot now that this is an example rather than a module with a test suite.

**Dropping `fail_closed`** was a good call. A single unconditional fail-closed path is easier to reason about than a flag that inverts the security posture, and it makes the earlier crash-vs-denial ambiguity harmless: every non-allow outcome blocks, whatever produced it. The reason string can still read a little oddly if the binary itself crashes (you get stderr text rather than "guard unavailable"), but nothing unsafe follows from it, and for an example that's fine.

CI is green on `cc261ddf` — and worth noting it's real signal here, not a no-op: both `ruff` and `pyrefly` include `examples/**/*.py` in their `sources`, and pyrefly only excludes `examples/playground/` and `examples/notebooks/`. So this file is genuinely linted and type-checked.

Approving on the technical merits. I'm leaving the actual merge to @Tomas2D for one reason that isn't about your code: this would be the first example in the repo built around a specific named third-party product, and the same question is live on #1644. That's a project-direction call for a maintainer to make once, consistently, rather than something I should settle per-PR.

PR #1622, opened 2026-08-29
feat(adapters): add OrcaRouter provider for Python and TypeScript

martinzudergaming-a11y:
## Summary

Adds [OrcaRouter](https://www.orcarouter.ai) as a first-class backend provider, mirroring the existing `MiniMax` adapter so it can be used through the same unified `ChatModel` interface and `ChatModel.from_name` factory.

OrcaRouter is an OpenAI-compatible AI gateway built for both models and agents. Like OpenRouter, it exposes a provider/model namespace across many models — but it also combines adaptive routing, automatic failover, zero-markup inference, observability, guardrails, and agent-tool governance behind the same endpoint. Adding `orcarouter` as a first-class provider means this project's users can use that stack directly, without treating OrcaRouter as an anonymous custom base URL. It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

### Changes

- **Python**: new `beeai_framework.adapters.orcarouter` package with `OrcaRouterChatModel` (backed by LiteLLM's OpenAI-compatible path), registered in `BackendProviders` (`orcarouter` alias). Default model is `orcarouter/auto`, base URL `https://api.orcarouter.ai/v1`, key from `ORCAROUTER_API_KEY`.
- **TypeScript**: new `typescript/src/adapters/orcarouter` adapter (`OrcaRouterChatModel` + `OrcaRouterClient` on `@ai-sdk/openai`), registered in the TS `BackendProviders`.
- **Examples**: `python/examples/backend/providers/orcarouter.py` and `typescript/examples/backend/providers/orcarouter.ts` following the existing provider example layout.
- **Tests**: unit/registration tests for Python (`tests/adapters/orcarouter/`) and TypeScript (`tests/e2e/adapters/orcarouter.test.ts`).
- **Docs**: added OrcaRouter to the supported-providers table in `docs/src/content/docs/modules/backend.mdx`.

### Verification

- Python: `ruff check`, `ruff format --check`, `pyrefly check` clean; 198 adapter/backend/agents/utils tests pass (incl. 14 new OrcaRouter tests).
- TypeScript: `tsc --noEmit`, `eslint`, `prettier --check` clean; 306 unit + 5 new OrcaRouter e2e tests pass.
- Live: `ChatModel.from_name("orcarouter:deepseek/deepseek-v4-flash")` and streaming via `orcarouter/auto` both returned successful responses against the OrcaRouter API.

### Notes

Models are addressed in OrcaRouter's namespaced format (e.g. `orcarouter/auto`, `deepseek/deepseek-v4-flash`). I'm an engineer on the OrcaRouter team.

Discord: discord.gg/YEubt8enRA · X: https://x.com/OrcaRouter

### DIFF
--- a/docs/src/content/docs/modules/backend.mdx
+++ b/docs/src/content/docs/modules/backend.mdx
@@ -38,6 +38,7 @@ The following table depicts supported providers. Each provider requires specific
 | MistralAI      |  ✅  | ✅        | MISTRALAI_CHAT_MODEL<br/>MISTRALAI_EMBEDDING_MODEL<br />**MISTRALAI_API_KEY**<br />MISTRALAI_API_BASE |
 | Transformers   |  ✅  | ✅        | TRANSFORMERS_CHAT_MODEL<br/>HF_TOKEN|
 | MiniMax        |  ✅  | ❌        | MINIMAX_CHAT_MODEL<br/>**MINIMAX_API_KEY**<br/>MINIMAX_API_BASE<br/>MINIMAX_API_HEADERS |
+| OrcaRouter     |  ✅  | ❌        | ORCAROUTER_CHAT_MODEL<br/>**ORCAROUTER_API_KEY**<br/>ORCAROUTER_API_BASE<br/>ORCAROUTER_API_HEADERS |
 
 <Tip>
 If you don't see your provider raise an issue [here](https://github.com/i-am-bee/beeai-framework/issues).
--- a/python/beeai_framework/adapters/orcarouter/__init__.py
+++ b/python/beeai_framework/adapters/orcarouter/__init__.py
@@ -1,0 +1,6 @@
+# Copyright 2025 © BeeAI a Series of LF Projects, LLC
+# SPDX-License-Identifier: Apache-2.0
+
+from beeai_framework.adapters.orcarouter.backend.chat import OrcaRouterChatModel
+
+__all__ = ["OrcaRouterChatModel"]
--- a/python/beeai_framework/adapters/orcarouter/backend/__init__.py
+++ b/python/beeai_framework/adapters/orcarouter/backend/__init__.py
@@ -1,0 +1,2 @@
+# Copyright 2025 © BeeAI a Series of LF Projects, LLC
+# SPDX-License-Identifier: Apache-2.0
--- a/python/beeai_framework/adapters/orcarouter/backend/chat.py
+++ b/python/beeai_framework/adapters/orcarouter/backend/chat.py
@@ -1,0 +1,58 @@
+# Copyright 2025 © BeeAI a Series of LF Projects, LLC
+# SPDX-License-Identifier: Apache-2.0
+
+import os
+
+from typing_extensions import Unpack
+
+from beeai_framework.adapters.litellm import LiteLLMChatModel, utils
+from beeai_framework.backend.chat import ChatModelKwargs
+from beeai_framework.backend.constants import ProviderName
+from beeai_framework.logger import Logger
+
+logger = Logger(__name__)
+
+ORCAROUTER_API_BASE = "https://api.orcarouter.ai/v1"
+
+
+class OrcaRouterChatModel(LiteLLMChatModel):
+    """
+    A chat model implementation for the OrcaRouter provider, leveraging LiteLLM.
+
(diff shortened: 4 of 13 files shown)


Tomas2D (commented):
Thanks for the thorough submission. To be clear up front: **the code is good.** It's a faithful port of the MiniMax adapter, the Python and TypeScript sides are consistent, the test coverage is real, and I found nothing objectionable in the implementation — no telemetry, no extra headers injected behind the user's back, no changes to workflows or build scripts. If we decide to take this, it's close to mergeable.

My hesitation is about whether we should take it at all, and that's a maintainer call rather than something you can fix in the diff.

**1. We don't make gateways first-class, including much larger ones.** OpenRouter — same category, vastly bigger catalog and userbase — is *not* a first-class provider here. It appears only as an enum value in the AgentStack adapter, mapped to `None`. Adding OrcaRouter while OpenRouter isn't in `BackendProviders` is hard to justify on any principle except who happened to open a PR.

**2. An OpenAI-compatible gateway already works today, in three lines:**

```python
OpenAIChatModel("orcarouter/auto", api_key=..., base_url="https://api.orcarouter.ai/v1")
```

That's the documented path for OpenAI-compatible hosts (see #1618). Against that, this PR is +532 lines and a permanent public API surface — `ProviderName`, `BackendProviders`, an alias, four env vars, a docs table row, in two languages — that we'd own and keep working indefinitely. The marginal user benefit over `base_url` is the `orcarouter:` prefix.

**3. Listing a provider in our docs table is an implicit endorsement,** which means we need to be able to verify the vendor independently. Right now I can't. The PR states you're an engineer on the OrcaRouter team, but the account opening it was created on 26 August, three days before this PR, with no other activity. That's not an accusation — new accounts are legitimate and vendor-authored adapters are normal and welcome. It does mean the affiliation needs to be established some other way before we put the name in front of our users. A note from a domain-verified address, or the PR being opened from an account tied to the company, would settle it.

Please also strip the marketing from the PR description if you revise it. "Zero-markup inference", "zero-trust security", "guardrails and agent-tool governance", plus Discord and X links, aren't claims we can evaluate or that belong in the commit record. What a reviewer needs is: what does this do that `base_url` doesn't?

Some smaller notes for whichever way this goes:
- `test_model_from_env` asserts on `anthropic/claude-opus-4.7`, which isn't a real model id. Harmless as a fixture string, but it does undercut the "verified live against the API" claim in the description — please use ids that exist in the catalog.
- I have **not** approved the gated CI run yet; no point spending it until the question above is settled. Happy to approve it immediately if we decide to proceed.

@Tomas2D to decide on the policy question. If the answer is "gateways stay non-first-class", the right outcome is probably a short docs section showing OpenAI-compatible gateways in general via `base_url`, which would serve OrcaRouter, OpenRouter and everyone else equally — and I'd be glad to take a PR for that from you.

Tomas2D:
Thanks for your patience while we settled the policy question, and thanks again for the quality of the submission — as I said in the review, the adapter itself is well built and I found nothing wrong with the implementation.

We've decided not to take gateways as first-class providers, so I'm going to close this. The reasoning isn't about OrcaRouter specifically:

- OpenRouter, the largest gateway in this category, isn't a first-class provider here either. Adding one gateway while the category leader isn't in `BackendProviders` would leave us with a support matrix determined by who happened to open a PR rather than by any principle.
- An OpenAI-compatible gateway already works today without any adapter, via `OpenAIChatModel(..., base_url=...)`. The marginal gain from this PR is the `orcarouter:` prefix, against a permanent two-language public API surface — `ProviderName`, `BackendProviders`, an alias, four env vars, a docs table row — that we'd own indefinitely.
- Listing a provider in our docs table reads as an endorsement, and we'd be committing to keep that integration working for every future user who picks it up from our docs.

What I'd genuinely welcome instead: a docs section on **using OpenAI-compatible gateways in general** via `base_url`, with a short worked example. That serves OrcaRouter, OpenRouter and everyone else equally, it's the thing our users are actually missing today, and it wouldn't be gated on any of the questions above. If you'd like to send that, I'll review it promptly — and using OrcaRouter as the worked example would be entirely fair.

Closing this one. Not a reflection on the work, and no objection to you or anyone else opening the docs PR.

PR #1640, opened 2026-09-07
feat(tools): add recoverable timeout option for tool runs

DaliborKr:
#### Problem

There was no built-in way to set a per-tool execution timeout. The only workaround was passing `AbortSignal.timeout()` via `ToolRunOptions.signal`, but that raises a fatal `AbortError` that kills the entire agent run workflow (agents can not catch it and continue).

#### Solution

Adds `timeout: float | None` to `ToolRunOptions`. When the tool runs too long, `asyncio.timeout()` interrupts it and raises a regular `ToolError` instead. The agent catches it, feeds it back to the model, and keeps going.

A few things to know:
- Timeout is per-attempt (each retry gets a fresh deadline)
- External `AbortSignal` aborts still work the same (setting `timeout` does not change the abort behavior)
- Opt-in — omitting `timeout` keeps existing behavior unchanged

#### Limitation

The timeout only works when the tool is waiting at an `async` suspension point. Blocking code like `time.sleep()` or `requests.get()` without `await` will not be interrupted. This is a shared limitation with the `AbortSignal.timeout()` as far as I know.

#### General

#### Code quality checks

#### Testing

#### Documentation
- [x] Documentation is updated

### DIFF
--- a/docs/src/content/docs/modules/tools.mdx
+++ b/docs/src/content/docs/modules/tools.mdx
@@ -162,6 +162,20 @@ console.log(result.getTextContent()); // prints data as text
 ```
 </CodeGroup>
 
+### Execution timeout
+
+Pass `timeout` (in seconds) to `ToolRunOptions` to limit how long a single tool call is allowed to run. When the deadline is exceeded, a recoverable `ToolError` is raised (the agent catches it, feeds it back to the model, and continues).
+
+```py Python
+result = await tool.run(input, options=ToolRunOptions(timeout=10.0))
+```
+
+The timeout is **per-attempt**: each retry gets a fresh deadline, so `timeout=5` with `max_retries=3` bounds each individual attempt at 5 s, not the total run.
+
+<Note>
+The timeout can only interrupt the tool at `async` suspension points. Blocking calls such as `time.sleep()` or `requests.get()` will not be interrupted, when not using `await`.
+</Note>
+
 ### Using tools with agents
 
 The true power of tools emerges when integrating them with agents. Tools extend the agent's capabilities, allowing it to perform actions beyond text generation:
--- a/python/beeai_framework/tools/tool.py
+++ b/python/beeai_framework/tools/tool.py
@@ -1,6 +1,7 @@
 # Copyright 2025 © BeeAI a Series of LF Projects, LLC
 # SPDX-License-Identifier: Apache-2.0
 
+import asyncio
 import contextlib
 import inspect
 import typing
@@ -103,6 +104,7 @@ class Tool(Generic[TInput, TRunOptions, TOutput], ABC):
         options_dict = options.model_dump(exclude_none=True) if options else {}
         options_dict.pop("signal", None)
         options_dict.pop("retry_options", None)
+        options_dict.pop("timeout", None)
         return BaseCache.generate_key(input, options_dict)
 
     async def clear_cache(self) -> None:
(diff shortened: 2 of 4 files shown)


Tomas2D (commented):
Nice PR — the problem statement is exactly right (`AbortSignal.timeout()` raising a fatal `AbortError` that the agent can't recover from is a real gap), the implementation is 13 lines, and it's opt-in. I approved the gated CI run and it's green.

The tests are the strongest part. `test_cache_hit_bypasses_timeout_and_excludes_timeout_from_cache_key` in particular tests something easy to get wrong and easy to forget — dropping `timeout` from `_generate_key` alongside `signal` and `retry_options` is exactly right, otherwise two identical calls with different deadlines would miss the cache. `test_timeout_retried_gets_fresh_deadline_per_attempt` and the abort-independence test are also well chosen.

## One thing I'd like changed

`except TimeoutError` catches more than your own deadline:

```python
async with asyncio.timeout(run_timeout):
    result = await self._run(validated_input, options, context)
except TimeoutError as e:
    raise ToolError(f"Tool '{self.name}' timed out after {run_timeout}s", cause=e) from e
```

If `_run` itself raises `TimeoutError`, it gets relabelled as *"Tool 'x' timed out after 30.0s"* even when the tool actually failed in 200ms against its own much shorter deadline. That's actively misleading when debugging. It's easy to hit: `socket.timeout` has been an alias of the builtin `TimeoutError` since 3.10, so any tool doing socket-level I/O, or using `asyncio.wait_for` internally, surfaces one.

`asyncio.timeout` gives you the discriminator directly:

```python
try:
    async with asyncio.timeout(run_timeout) as cm:
        result = await self._run(validated_input, options, context)
except TimeoutError as e:
    if not cm.expired():
        raise  # the tool's own timeout, not ours
    raise ToolError(f"Tool '{self.name}' timed out after {run_timeout}s", cause=e) from e
```

Worth a test with a tool that raises `TimeoutError` promptly under a long `timeout`, asserting the original propagates rather than being rewritten.

## Two smaller points

**Total wall-clock isn't obvious.** `ToolError` is non-fatal and retryable by default, so `timeout=5` with `retry_options(max_retries=3)` bounds each attempt at 5s but the whole run at ~20s. Your per-attempt choice is right and you tested it — I'd just say so on the field, since `timeout` currently has no description:

```python
timeout: float | None = Field(default=None, description="Per-attempt execution deadline in seconds. Each retry gets a fresh deadline.")
```

The `Limitation` note from your description (only interrupts at `async` suspension points) deserves to be in that docstring too — it's the thing people will trip on.

**Docs.** You left "Documentation is updated" unchecked. `ToolRunOptions` is public API and appears throughout `docs/src/content/docs/modules/tools.mdx`; a short subsection on the timeout option would be welcome in this PR.

## Question, not a request

`BaseToolRunOptions` exists on the TypeScript side too (`typescript/src/tools/base.ts:53`). Is Python-only intentional here? Perfectly fine if so — plenty of features land Python-first — but if you think parity is wanted I'd rather open a tracking issue now than discover the divergence later.

Fix the `cm.expired()` case and I'm happy to merge this.

DaliborKr:
Thanks you for a quick review. Your points are valid indeed and I tried to resolve them all.

Regarding the Python-only implementation note. Yes, Python-only is intentional. I haven't looked into whether TypeScript has the same gap, but I'm happy to open a tracking issue so it can be investigated separately.

Tomas2D:
Merged — thanks, this was a pleasure to review. You addressed every point, and the `timeout_ctx.expired()` handling plus `test_tool_internal_timeout_error_propagates_unchanged` is exactly the shape I was hoping for: asserting the *absence* of the misleading message is a better test than asserting its presence would have been.

Yes please on the TypeScript tracking issue — worth noting there that `BaseToolRunOptions` is at `typescript/src/tools/base.ts:53`, and that the TS gap may differ since its abort handling isn't a direct mirror of the Python side. No commitment implied to implementing it; having the divergence recorded is the valuable part.

PR #1644, opened 2026-09-08
feat(agents): show a self-contained preflight requirement

AgentTanuki:

No separate issue. This example shows how a caller's endpoint-evidence policy can restrict a selected MCP tool without granting execution authority or replacing other requirements.

### Description

Adds a self-contained example under `python/examples/agents/requirement/` and offline tests. A native `Requirement` contributes refusal rules and obtains a fresh observation at the blocking tool-start boundary before the exact selected MCP tool can run. The caller approves endpoint disclosure, selects the model, task and tool, and decides which unknown observations to tolerate. Evidence never supplies execution authority.

`--preflight-url` selects a compatible HTTPS observation service; the default is Agent Guild's free `/preflight`. The selected URL is validated before opening the MCP transport and preserved when the requirement is cloned. Disclosure approval applies to the selected service and its data handling; Agent Guild probes and logs the submitted URL. Neither target nor observation-service URLs may contain credentials, queries or fragments. Requests have a five-second total deadline and 64 KiB response limit, with no redirects, environment proxies, credentials or payment attempts.

The example requires all six current check names and consistent failed, unknown and scored summaries. Additional checks obey the same caller policy. Failed, unavailable, invalid and unapproved unknown evidence blocks execution. Required reachability and handshake checks cannot be waived by tolerating unknowns. Retained unsigned bytes are untrusted evidence: they do not establish verified identity, signature validity, completed work, payment binding or safety. MCP initialization/discovery precedes the guarded `tools/call` boundary.

The two-file contribution uses existing framework dependencies and direct imports, with no separate package or core documentation section. Copyright headers follow the repository's 2025 convention.

### Validation

On Python 3.12 with the repository's locked tooling and native framework 0.1.84, **49 focused offline tests pass** (47 example tests and two existing requirement tests). Real `ClientSession`, `MCPTool` and `RequirementAgent` paths use an in-memory HTTP transport and deterministic model; network connections are blocked. Coverage includes default/custom service selection, CLI forwarding, cloning, changed evidence at execution, unknown and omitted checks, response consistency, failure handling and another requirement's refusal.

Locked Ruff check/format, Pyrefly and CLI help pass. Required snippet regeneration and verification pass for 28 MDX files and 148 directives without generated changes. Full local monorepo tests, provider E2E and a documentation site build were not run; GitHub checks report validation for their associated commit.

#### General
- [x] I have read the Python contributor guide and Code of Conduct.
- [x] I have signed off on the revision commit.
- [x] Commit messages and PR title follow the project's conventions.
- [x] The existing PR has the `python` label.

#### Code quality checks
- [ ] Full local `mise check` passes. Focused locked Ruff and Pyrefly checks pass for the example and its tests.

#### Testing
- [ ] Full local upstream unit suite passes.
- [ ] Provider E2E suite passes.
- [x] All 49 focused offline tests pass without a live model, MCP service or Guild request.

#### Documentation
- [x] Usage and disclosure limits are in the example docstring; required snippet regeneration and verification pass.

### DIFF
--- a/docs/src/content/docs/modules/agents/requirement-agent.mdx
+++ b/docs/src/content/docs/modules/agents/requirement-agent.mdx
@@ -587,6 +587,133 @@ for (const prompt of prompts) {
 
 </CodeGroup>
 
+## Endpoint preflight before remote tool execution (Python)
+
+The optional [Agent Guild integration](https://github.com/AgentTanuki/beeai-agentguild/tree/8a0bf774d68e54bf4cf3a79bf0e9d65bff8f76f7) demonstrates a custom requirement that applies caller-selected policy before native MCP tool execution. It uses `RequirementAgent`, actual `MCPTool` instances, and the blocking tool-start mechanism also used by `AskPermissionRequirement`. Deny-only rules constrain the model's offered tools; the tool-start hook preserves the restriction through the released agent's invalid-final-answer fallback and refreshes expired observations before execution.
+
+Install the optional source in a separate environment. This example is validated with `beeai-framework==0.1.83`; it adds no dependency to the framework's base installation and does not require an unpublished PyPI package:
+
+```sh
+python -m venv .venv
+. .venv/bin/activate
+python -m pip install 'beeai-agentguild @ git+https://github.com/AgentTanuki/beeai-agentguild.git@8a0bf774d68e54bf4cf3a79bf0e9d65bff8f76f7'
+```
+
+Running the example discloses the caller-selected public endpoint to `https://agent-guild-5d5r.onrender.com/preflight`, which performs a bounded observation. The caller must authorize that disclosure, the configured model, and any later task execution. Preflight receives the endpoint URL, not the task, MCP credentials or model prompt. Credential-bearing URLs, query strings, fragments and local/private literal targets are rejected. This example supplies no authentication; initialization and tool discovery occur before the requirement, while subsequent `tools/call` invocations are guarded. Only tool names explicitly selected by the caller are exposed.
+
+<CodeGroup>
+{/* <!-- embedme python/examples/agents/requirement/agent_guild_preflight.py --> */}
+```py Python [expandable]
+"""Apply caller-selected preflight policy before native MCP tool execution.
+
+Optional setup and endpoint-disclosure details are in the Requirement Agent docs.
+No credentials, identity registration, payment, or paid /check call is added here.
+"""
+
+import argparse
+import asyncio
+
+try:
+    # pyrefly: ignore [missing-import]
+    from beeai_agentguild import (
+        AgentGuildPreflightRequirement,
+        Observation,
+        PreflightClient,
+        PreflightPolicy,
+        RemoteToolBinding,
+        mcp_bindings,
+    )
+except ImportError as error:
+    raise ImportError("Install the optional pinned beeai-agentguild source from the Requirement Agent docs.") from error
+
+from beeai_framework.agents.requirement import RequirementAgent, RequirementAgentOutput
+from beeai_framework.backend import ChatModel
+
+
+async def delegate(
+    *,
+    model: ChatModel,
+    bindings: tuple[RemoteToolBinding, ...],
+    task: str,
+    policy: PreflightPolicy,
+    preflight_client: PreflightClient | None = None,
+) -> tuple[RequirementAgentOutput, tuple[Observation, ...]]:
+    # Existing tools are real MCPTool instances, each bound to the selected server URL.
+    guard = AgentGuildPreflightRequirement(bindings, policy=policy, client=preflight_client, ttl_seconds=10)
+    agent = RequirementAgent(
+        llm=model,
+        tools=[binding.tool for binding in bindings],
+        requirements=[guard],
+    )
+    result = await agent.run(task)
+    return result, guard.evidence
+
+
+async def main() -> None:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("--model", required=True, help="Caller-configured BeeAI model, such as ollama:granite4:micro")
+    parser.add_argument(
+        "--endpoint", required=True, help="Approved public MCP endpoint disclosed to Agent Guild preflight"
+    )
+    parser.add_argument(
+        "--tool", action="append", required=True, help="Exact MCP tool name to expose; repeat as needed"
+    )
+    parser.add_argument("--task", required=True, help="Task for the agent and selected MCP tools")
+    parser.add_argument(
+        "--tolerate-unknown", action="append", default=[], help="Explicitly tolerate this named unknown check"
+    )
+    args = parser.parse_args()
+
+    # Keep both blocking checks required. Caution and all unlisted unknowns block.
+    # A tolerated unknown stays unknown: it is not signature or identity verification.
+    policy = PreflightPolicy(tolerated_unknowns=frozenset(args.tolerate_unknown))
+    async with mcp_bindings(args.endpoint, include=set(args.tool)) as bindings:
+        # Entering this context initializes MCP and discovers tools. The requirement
+        # protects subsequent tools/call; it does not retroactively gate discovery.
+        result, evidence = await delegate(
+            model=ChatModel.from_name(args.model), bindings=bindings, task=args.task, policy=policy
+        )
+
+    print(result.last_message.text)
+    for observation in evidence:
+        print(
+            {
+                "source_origin": observation.source_origin,
+                "synthetic": observation.synthetic,
+                "retrieved_at": observation.retrieved_at,
+                "status": observation.status,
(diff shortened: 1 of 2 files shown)


Tomas2D:
Still a draft, so this isn't a blocking review — but the direction has some structural problems that are worth settling before more effort goes in, since they affect where (and whether) this lands rather than how the code reads.

Credit where due first: the writeup is unusually careful about scope. Explicitly stating that preflight is "a **free unsigned observation**, not signature validation, verified identity, complete protocol conformance, successful task execution, payment binding or a safety guarantee", disclosing exactly what gets sent to the endpoint, and noting that `observation.raw_body` is untrusted evidence rather than instructions — that's the right instinct, and more rigorous than most integration docs.

**1. It puts a specific third-party commercial service into a first-party docs page.** This adds 127 lines to [requirement-agent.mdx](https://github.com/i-am-bee/beeai-framework/blob/main/docs/src/content/docs/modules/agents/requirement-agent.mdx), the canonical page for a core framework feature, and hardcodes `https://agent-guild-5d5r.onrender.com/preflight`. Two issues with that:

- No other page under `docs/src/content/docs/` references an `onrender.com` host — I checked. The external services the docs do name are LLM providers, and they're named as interchangeable options, not as a single endorsed vendor.
- The service has a paid tier. The docs are careful to say this path only calls free `/preflight` and "never invokes paid `/check`", which I believe — but a reader landing on the core Requirement Agent page still encounters a funnel to a commercial product, and the framework takes on implicit responsibility for its availability and behavior.

This is the same scope question raised on #1610 (HOL Guard), and it needs a maintainer's answer rather than mine. My own read is that the two should be resolved the same way, and consistently.

**2. It tells users to `pip install` from a pinned commit on a personal repo.**

```sh
python -m pip install 'beeai-agentguild @ git+https://github.com/AgentTanuki/beeai-agentguild.git@8a0bf774...'
```

Official docs instructing an install of unreleased code from an individual's GitHub, at a commit that the framework doesn't control and can't audit on an ongoing basis, is a meaningfully different ask than `pip install` from PyPI. At minimum this should wait for a published package.

**3. The pinned version reference is already stale.** The docs say "validated with `beeai-framework==0.1.83`"; `python_v0.1.84` shipped yesterday. Pinning a framework version inside framework docs guarantees this drifts — and since the integration lives outside this repo, nothing in CI will catch it when it does.

**4. Nothing here is verifiable by CI.** The example guards its imports with `try/except ImportError`, so `beeai_agentguild` is never installed in CI and the code path is never exercised. The "four deterministic loopback executions" you mention live in the external repo. So the framework would be publishing, on a first-party page, an example that it cannot prove still works — and the failure mode when it breaks is a user's `pip install` from a git URL.

**Concrete suggestion.** The mechanism you're demonstrating — a custom `Requirement` that applies caller policy at the blocking tool-start boundary before remote tool execution — is genuinely interesting and is exactly the sort of thing the requirement system is for. That mechanism could be documented generically, with a local/loopback policy check standing in for the service, and no external dependency or endpoint at all. Users learn the pattern, the example runs in CI, and Agent Guild remains linked from its own repo for those who want the hosted version. That version I'd expect to be uncontroversial.

Flagging @Tomas2D on the scope question, since it's the same call as #1610 and shouldn't be decided per-PR.

Tomas2D:
Update on the scope question I flagged above — it's been decided, so you're not waiting on it.

**Naming a specific third-party product in an example is acceptable.** #1610 was just merged on that basis: it landed as a single `python/examples/middleware/hol_guard.py`, after being reworked from a vendored adapter under `beeai_framework/adapters/` into a plain example. So point 1 of my review no longer blocks you in principle.

Two things that decision does *not* cover, though, and they're the ones that matter most here:

**Placement is still different from #1610.** That PR added a file under `python/examples/`. This one adds 127 lines to [requirement-agent.mdx](https://github.com/i-am-bee/beeai-framework/blob/main/docs/src/content/docs/modules/agents/requirement-agent.mdx) — the canonical docs page for a core framework feature — and hardcodes `https://agent-guild-5d5r.onrender.com/preflight` into it. "A vendor example may live in `examples/`" is a narrower ruling than "a vendor section may live in a core docs page." Following the #1610 shape as closely as possible is the path of least resistance: lead with the example file, keep any docs footprint minimal.

**Points 2–4 stand unchanged and are independent of the vendor question:**

- `pip install` from a pinned commit on a personal GitHub repo, rather than from PyPI.
- The `beeai-framework==0.1.83` pin is now two releases stale — `python_v0.1.84` shipped yesterday.
- Nothing here is exercised by CI. Worth knowing that this is a solved problem: both `ruff` and `pyrefly` include `examples/**/*.py` in their `sources` (pyrefly excludes only `examples/playground/` and `examples/notebooks/`), so an example file *is* linted and type-checked — which is exactly why #1610's green run was meaningful. Your `try/except ImportError` guard means the import path stays unverified regardless, but the file itself would at least be checked.

The generic-mechanism suggestion from my earlier comment is still the version I'd expect to sail through, but it's no longer the only option open to you.

Tomas2D:
This is a substantial improvement and it addresses the structural objections from my earlier reviews. Summarising where it now stands, since the PR has changed shape twice.

**Resolved:**

- **No docs-page footprint.** The 127-line section in `requirement-agent.mdx` is gone. That was my main concern after the vendor-example precedent was settled on #1610, because "a vendor example may live in `examples/`" is a narrower ruling than "a vendor section may live in a core docs page."
- **No external package.** The `pip install` from a pinned commit on a personal GitHub repo is gone; the example now imports only `httpx`, `mcp` and framework modules. That also removes the stale `beeai-framework==0.1.83` pin, which had drifted two releases.
- **It's now verifiable by CI**, which was point 4 of my first review and the one I cared most about.

**The tests are the strongest part of this.** `pytestmark = pytest.mark.unit` means they actually run under `pytest -m unit`, and the `offline` fixture monkeypatching `socket.connect`/`connect_ex` to raise is a genuinely good call — it doesn't just avoid the network, it *proves* the code path never reaches for it. Combined with `httpx.MockTransport`, you get real coverage of the branches that matter: connect failure, a 302 redirect, a 65537-byte oversized body, malformed and duplicate checks, and omitted checks not silently erasing unknowns. `unique_object` rejecting duplicate JSON keys is a nice touch most implementations miss.

I also appreciate that the module docstring keeps the scope disclaimers explicit — preflight as evidence rather than authorization, and retained bytes as untrusted data rather than instructions.

**Remaining, minor:**

1. `PREFLIGHT_URL` still hardcodes `https://agent-guild-5d5r.onrender.com/preflight`. Naming a third-party service in an example is now allowed (that was settled on #1610), so this is no longer blocking — but an `onrender.com` hostname is not a stable address, and when it stops resolving the example silently becomes a demonstration of the fail-closed path rather than of preflight. Worth at least making it overridable by a flag or env var so the example outlives the host.
2. The copyright header says `2026 ©`; every other file in the tree says `2025 ©`. `addlicense` checks for the pattern rather than the year so CI won't object, but it's inconsistent.

I've approved the workflow runs so you get real CI signal on this revision — both `ruff` and `pyrefly` include `examples/**/*.py`, so the example itself is linted and type-checked, not just the test.

Still marked as a draft, so I'm not treating this as ready to merge — flip it out of draft when you are and I'll take another pass.

Tomas2D (approved):
Both of the remaining nits from my last pass are addressed, so from my side this is in good shape.

- `--preflight-url` makes the observation service configurable, with the hosted instance now only a default. That was the substantive one: the example no longer silently becomes a demonstration of the fail-closed path if a particular `onrender.com` host stops resolving. Running `validate_endpoint()` over the supplied URL as well as the target is a nice touch — the flag can't be used to point the example at a credential-bearing or non-HTTPS endpoint.
- Copyright header now reads `2025 ©`, consistent with the rest of the tree.

For anyone picking this up cold, the shape has changed substantially since it was opened: it started as a 127-line section in the core `requirement-agent.mdx` plus an example depending on a package installed from a pinned commit on a personal GitHub repo. It is now a self-contained example plus tests, no docs-page footprint and no external package. The vendor-example question it originally raised was settled separately on #1610.

The tests remain the part I'd point at. `pytestmark = pytest.mark.unit` means they run under `pytest -m unit`, and the `offline` fixture that monkeypatches `socket.connect`/`connect_ex` to raise doesn't merely avoid the network — it proves the code path never reaches for it, which is the right assertion for an example that talks to a remote service. The `httpx.MockTransport` cases cover the branches that actually matter: connect failure, a 302 redirect, an oversized body, malformed and duplicate checks, and omitted checks not silently erasing unknowns.

I've approved the workflow runs now that this is out of draft, so you'll get real CI signal on the current head. Worth noting both `ruff` and `pyrefly` include `examples/**/*.py` in their `sources` (only `examples/playground/` and `examples/notebooks/` are excluded), so the example itself is linted and type-checked, not just the test file.

Leaving the merge decision to @Tomas2D, but I have no further requests on the code.

PR #1647, opened 2026-09-09
fix(backend): add default 600s request timeout for LiteLLM chat model

saitejagirada:
## Problem

`LiteLLMChatModel` called litellm's `acompletion()` without a `timeout` kwarg. When a backend accepts the connection but never responds, the call blocks for litellm's internal fallback of 600 seconds (or indefinitely), hanging the entire agent run.

## Solution

- Add `DEFAULT_REQUEST_TIMEOUT_SECONDS = 300` (configurable via `BEEAI_DEFAULT_REQUEST_TIMEOUT` env var)
- Apply `litellm_input.setdefault("timeout", DEFAULT_REQUEST_TIMEOUT_SECONDS)` in both `_create` (non-streaming) and `_create_stream` (streaming) paths
- Uses `setdefault` so user-supplied `settings={"timeout": <seconds>}` takes precedence

### Production code — `chat.py`
- New module-level constant `DEFAULT_REQUEST_TIMEOUT_SECONDS` with env var override support
- `setdefault` guard added to both `_create` and `_create_stream` methods

### Test code — `test_litellm_chat.py`
- `test_non_streaming_uses_default_timeout` — verifies default 300s is passed
- `test_streaming_uses_default_timeout` — same for streaming path
- `test_settings_override_takes_precedence` — verifies user override (42s) wins
- Function-scoped `timeout_model` fixture to avoid scope mismatch with `monkeypatch`

## Test Results

```
16 passed in 2.31s
```


Tomas2D (changes requested):
Good catch on the underlying gap — there's currently no timeout anywhere in `beeai_framework/backend/` or the LiteLLM adapter, so a non-responding backend really does hang the run. Using `setdefault` so `settings={"timeout": ...}` wins is the right shape, and the `BEEAI_*` env prefix matches the existing convention (`BEEAI_DDG_TOOL_PROXY`, `BEEAI_OPEN_METEO_TOOL_PROXY`).

**Blocker: DCO.** Commit `6a342a8` ("refactor: apply review suggestions for timeout PR") has no `Signed-off-by` trailer — `17360bb` does. `git rebase --signoff -x true HEAD~2` or an amend on that commit will clear the check.

**Bug — a bad env var crashes an unrelated import.**

```python
DEFAULT_REQUEST_TIMEOUT_SECONDS = int(os.getenv("BEEAI_DEFAULT_REQUEST_TIMEOUT", "300"))
```

`int()` raises `ValueError` on anything non-integral, and this runs at module import. `BEEAI_DEFAULT_REQUEST_TIMEOUT=300.5` — a perfectly reasonable thing for someone to set given the value is documented in seconds — takes down `import beeai_framework.adapters.litellm.chat`, and with it every provider that builds on it (openai, ollama, watsonx, anthropic, …). The traceback points at an import line, not at the env var, so it's a rough thing to debug. Suggest `float()` plus a guarded parse that falls back to the default and logs a warning on a bad value.

**Question for the maintainers, not a change request:** 300s is a hard default for *every* LiteLLM-backed call. Long reasoning generations can legitimately run past five minutes, so this converts some currently-slow-but-successful runs into timeouts. Worth confirming 300 is the number you want rather than something more permissive, since the failure mode is silent from the user's side — they just see a timeout where they previously got an answer.

**Minor:** the docstring says the value is overridable "globally via the `BEEAI_DEFAULT_REQUEST_TIMEOUT` env var", but it's read once at import, so setting it from Python after the module is imported has no effect. Worth saying "must be set before import".

**Tests** look solid — both paths plus the override, and the comment explaining why `fake_acompletion_stream` returns a generator rather than being one is genuinely helpful. Consider adding one for the bad-env-var path once that's guarded.

saitejagirada:
Thanks for the thorough review! I've addressed all the feedback Squashed into a single signed commit (DCO fix) , Replaced int() with a guarded float() parse that falls back to default + logs a warning on invalid values ,Clarified docstring: env var is read once at import time ,Added tests for the bad_env_var edge cases

Regarding the default timeout value: you raised a good point that 300s could be too aggressive for long reasoning generations. I'd love the team's guidance here — what default would you prefer?

May be we could change it to 600 default

Tomas2D (changes requested):
Thanks for turning this around so quickly — DCO is green now, and `_parse_timeout_env()` is a clean fix for the import-time crash. Rejecting zero/negative, falling back with a warning instead of raising, and using `float()` so `"300.5"` works all address exactly what I raised. The eight `TestParseTimeoutEnv` cases cover the branches properly.

Two things left, then this is good to go.

**1. CI is red — `ruff` import sorting** ([log](https://github.com/i-am-bee/beeai-framework/actions/runs/35047197491/job/104679478634)):

```
tests/backend/providers/test_litellm_chat.py:4:1: I001 [*] Import block is un-sorted or un-formatted
Would reformat: tests/backend/providers/test_litellm_chat.py
```

`mise fix` resolves both. Nothing wrong with the logic — just the new `from collections.abc import AsyncGenerator` landing out of order.

**2. Maintainer decision on the default: please change 300 → 600.**

I raised the default value as an open question and it's now been decided — the default should be **600s**, not 300s. The reasoning: 300s converts calls that currently *succeed* slowly into failures, and long reasoning / extended-thinking generations legitimately run past five minutes. That's a silent behavior change for those users on a version bump, and the symptom (a timeout where an answer used to appear) is hard to trace back to a framework default. 600s keeps everyone who works today working, while still guaranteeing the call terminates — which is the actual bug you found.

Concretely: `_DEFAULT_TIMEOUT_FALLBACK = 600.0`. Your tests reference the constant symbolically rather than hardcoding `300`, so they should all still pass unchanged — nice.

One knock-on: the comment above the constant currently reads *"Prevents agent runs from hanging for LiteLLM's internal fallback of 600 seconds (or indefinitely)"*, which becomes self-contradictory once the default **is** 600. Worth rewording so the rationale stays honest — something like: the framework should never depend on an upstream default it doesn't control, and the case this actually fixes is the connection that hangs indefinitely rather than falling back at all. Same for the PR title and commit body, which both say 300s.

**Nit, non-blocking:** `float("nan")` slips through the guard — `nan <= 0` is `False`, so a `BEEAI_DEFAULT_REQUEST_TIMEOUT=nan` would be passed to litellm as-is. `math.isfinite(value)` in the same condition closes it. Obscure enough that I wouldn't hold the PR for it, though `inf` arguably becomes a useful undocumented "no timeout" escape hatch if you'd rather allow that one deliberately.

Ping me when those are pushed and I'll get the workflow approved and merge.

saitejagirada:
@Tomas2D everything seems to fine Now

Tomas2D:
This revision addresses everything. Thanks for the quick turnarounds.

- `_DEFAULT_TIMEOUT_FALLBACK = 600.0` — matches the maintainer decision.
- `math.isfinite` is in the guard, so `BEEAI_DEFAULT_REQUEST_TIMEOUT=nan` and `=inf` now fall back rather than being passed through to litellm. That was a non-blocking nit in my last review and you picked it up anyway — the `test_inf_falls_back` / NaN cases are a nice addition.
- The constant's comment is correctly reworded. It no longer justifies the value by reference to litellm's own 600s fallback (which would have been circular now that the default *is* 600), and instead makes the real argument: the framework shouldn't depend on an upstream default it doesn't control, and the case actually being fixed is the connection that hangs rather than falling back at all.

**Two things I've taken care of directly**, since this had already been round-tripped on lint twice:

1. `ruff check` was clean but `ruff format --check` still failed on `tests/backend/providers/test_litellm_chat.py`. It was a single trailing blank line at end of file. I've pushed `6288e8a4` to your branch with the formatter's output — no functional change. For future reference, `mise fix` runs both `ruff check --fix` and `ruff format`, which is why `ruff check` alone kept coming back clean while CI failed.

2. The PR title still read *"add default 300s request timeout"* while the code does 600s. Since this repo squash-merges, that title becomes the commit subject on `main`, so I've corrected it to 600s. The PR description above still describes 300s in a couple of places — not worth another push, but flagging it so the text isn't mistaken for the current behaviour.

I've approved the workflow run for the new head. If it comes back green this is ready to merge.

For the record, verified locally against your branch: `ruff check` clean, `ruff format --check` clean over all 608 files, and 25/25 passing in `test_litellm_chat.py`.

PR #1649, opened 2026-09-11
fix(agents): use union instead of intersection for ConditionalRequirement target validation

mittalpk:
While not required for a fix this small, no existing issue covers this — found via code review.

`ConditionalRequirement.init()` builds `targets` as the intersection of `_before`, `_after`, `_force_after`, and `{source}` before passing it to `_assert_all_rules_found()`, which checks that every name in `targets` corresponds to an actual tool. But `_check_invariant()` already enforces that `_before`, `_after`, and `_force_after` are pairwise disjoint (raising ValueError otherwise), so their intersection is always empty by construction — the validation is a permanent no-op regardless of what's passed to `only_before`, `only_after`, `force_after`, or `target`.

Concretely: `ConditionalRequirement(target=some_tool, only_before=["typo_tool_name"])` passes `init()` silently even when no tool named `typo_tool_name` exists, instead of raising the intended "Tool 'typo_tool_name' is specified as ... but not found."

Fixed by using union, so every referenced name is actually checked.

### How did you test it
Added `tests/agents/test_conditional_requirement.py` (no prior test coverage existed for this class): one test confirming a typo'd `only_before` entry now raises `ValueError`, one confirming a valid entry still passes cleanly. Verified both fail/pass correctly against pre-fix code via `git stash`. Full unit suite: 339 passed, 1 skipped (pre-existing, unrelated to this change), same as before this diff. `ruff check`, `ruff format --check`, and `pyrefly check` all clean.

### DIFF
--- a/python/beeai_framework/agents/requirement/requirements/conditional.py
+++ b/python/beeai_framework/agents/requirement/requirements/conditional.py
@@ -102,7 +102,7 @@ class ConditionalRequirement(Generic[TInput], Requirement[TInput]):
     async def init(self, *, tools: list[AnyTool], ctx: RunContext) -> None:
         await super().init(tools=tools, ctx=ctx)
 
-        targets = self._before & self._after & self._force_after & {self.source}
+        targets = self._before | self._after | self._force_after | {self.source}
         _assert_all_rules_found(targets, tools)
 
         for tool in tools:
--- a/python/tests/agents/test_conditional_requirement.py
+++ b/python/tests/agents/test_conditional_requirement.py
@@ -1,0 +1,39 @@
+# Copyright 2025 © BeeAI a Series of LF Projects, LLC
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+
+from beeai_framework.agents.requirement.requirements.conditional import ConditionalRequirement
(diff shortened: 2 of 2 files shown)


Tomas2D (approved):
Confirmed the analysis. `_check_invariant()` rejects `_before & _after` and `_before & _force_after`, so the four-way intersection could only ever be non-empty if a name appeared in `_after`, `_force_after` *and* `{source}` simultaneously — and `{self.source}` is a single element, so in practice the set was always empty and `_assert_all_rules_found` was a no-op. Union is the right fix.

It also lines up with how the rest of the module behaves: `ask_permission.py` calls `_assert_all_rules_found` on `_include`/`_exclude` unconditionally, and the `@requirement` decorator does the same on its targets. This was the odd one out.

One thing for the changelog rather than the diff: this is a user-visible behavior change. Configs that reference a tool name not present in the agent's tool list used to construct silently and now raise `ValueError` at `init()`. That's the intended behavior and it surfaces real typos, but someone will hit it on upgrade.

Tests are appropriately scoped — one failing case, one passing case, both marked `unit` so they actually run under `pytest -m unit`. CI green.

PR #1659, opened 2026-09-17
fix(internals): prevent extra AbortError after handled tool failures

DaliborKr:
### Description
When a BeeAI tool fails with a recoverable `ToolError`, the context was destroyed while its internal `abort_task` was still waiting for a signal. Destroying the context then woke this task and let it raise an `AbortError("Context has been destroyed.")` on its own, rather than through the usual `Task.cancel()` path.

This did not break the workflow, as the agent could handle the original `ToolError` and continue normally. However, `AbortError` is also a plain `Exception`, so tools like Sentry picked it up as a second, unhandled error next to the original `ToolError`.

This change cancels and waits for pending internal tasks before checking their result. Abort behavior stays the same, except for two things:

  - `abort_task` is now always cancelled through `Task.cancel()` instead of being left to raise `AbortError` on its own
  - if both tasks happen to finish in the same tick, abort now takes priority over a same-tick result, which it did not before.

#### General

#### Code quality checks

#### Testing

#### Documentation


Tomas2D (approved):
Traced this through and the diagnosis holds exactly. Approving.

The mechanism, for the record: `abort_task` awaits a future that `_on_abort` completes with `AbortError(context.signal.reason)`. In the old ordering, `runner_task.result()` was called *before* `abort_task.cancel()`, so when the runner had failed with a recoverable `ToolError` the `.result()` raised and the cancel line was never reached. Control then went to the `finally`, which calls `context.destroy()` → `self._controller.abort("Context has been destroyed.")` ([context.py:180-182](https://github.com/i-am-bee/beeai-framework/blob/main/python/beeai_framework/context.py#L180-L182)) → `_on_abort` sets `AbortError` on the future of a task nobody ever retrieves. Hence the phantom second error in Sentry alongside the real `ToolError`.

Cancelling everything still pending *before* touching any `.result()` is the right shape, and it makes the ordering obvious rather than incidental.

**On the same-tick priority change**, which you flagged yourself: when both tasks land in `done` together, abort now wins where the runner used to. I think that's the correct choice — an abort that has already fired shouldn't be silently discarded because a result happened to arrive in the same iteration — but it is a real semantic change and worth a line in the changelog, because the observable effect is an `AbortError` where a caller previously got a value. Note the `success` event is also skipped in that case, which is consistent.

The other paths check out: runner-done/abort-pending cancels and awaits `abort_task`, then `abort_task in done` is false so the result is returned normally; abort-done/runner-pending cancels the runner and raises. `gather(..., return_exceptions=True)` over the cancelled set is unchanged from before.

**One request: this needs a regression test.** The checklist has "Tests are included" unticked, and this is core control flow in `context.py` where the failure mode is a *stray* exception rather than a wrong return value — precisely the kind of thing that silently regresses. Something along the lines of: run a tool that raises a recoverable `ToolError` inside a `RunContext`, and assert no second `AbortError` surfaces — e.g. by installing an `asyncio` exception handler or checking `abort_task` was cancelled rather than completed. Without it, a future refactor that reorders these lines reintroduces the bug with a green suite.

CI is already green on your head commit.

DaliborKr:
Thanks for the feedback.

I already added the test that checks whether `abort_task` was actually cancelled through `Task.cancel()`, instead of just checking `abort_task.cancelled()`. That check alone isn't enough, since `AbortError` is also a `CancelledError` subclass, so `.cancelled()` reads `True` either way: whether the task was cancelled or left to raise `AbortError` on its own. I verified the test fails against the old code and passes against the fix.

On the changelog note: I noticed that the changelog is auto-generated here, so I wanted to ask for advice on how I should properly record the behavioral change there.

Tomas2D (approved):
Test added — thanks, and it's a better test than I had in mind.

I suggested asserting that no stray `AbortError` surfaces, which is awkward to observe reliably. Hooking `asyncio.create_task`, singling out the task named `abort-task` and wrapping its `.cancel` instead asserts the actual invariant: that the task is disposed of *through* `Task.cancel()` rather than being left to resolve its own `AbortError` when `destroy()` fires the signal. `assert abort_task in cancel_calls` is the discriminating line — `assert abort_task.done()` alone would pass either way, since pre-fix the task was also "done", just done by raising.

Since I asked for the test, I verified it isn't passing vacuously. Reverting `context.py` to the previous `if runner_task in done: / else:` shape and re-running:

```
FAILED python/tests/test_context.py::test_recoverable_tool_error_does_not_leave_abort_task_dangling
1 failed
```

and with your version restored it passes. So it genuinely pins the fix.

Also ran the full unit suite on your branch: **364 passed** (`pytest -m unit`, excluding `tests/adapters/mcp`, which doesn't collect against my stale local `mcp` — unrelated to this change).

The docstring explaining *why* the assertion is shaped that way is worth having; this is exactly the kind of invariant that a future refactor would quietly undo.

One thing from my earlier review still stands, and it's for the changelog rather than the diff: when both tasks complete in the same tick, abort now takes priority where the runner previously won, so a caller can observe an `AbortError` where they used to get a value. Correct behaviour in my view, but user-visible.

CI green. Good to merge from my side.

PR #1660, opened 2026-09-17
fix(tools): preserve leading-hyphen grep patterns

baba9811:

`GrepTool` passes its regex pattern directly to ripgrep without terminating option parsing. Searching a file containing `usage: app --help` with `pattern="--help"` therefore runs ripgrep's help command and returns an empty match list. `--version` also produces empty results, while `-needle` is parsed as options and can return the shortened match `edle`. The standard-library fallback searches these patterns correctly.

### Description

Insert `--` before the pattern and search root so ripgrep treats them as positional arguments. Add a parameterized regression test through the public `GrepTool.run` API using real ripgrep for all three patterns, checking the file, line, text and complete match.

### Testing

- All three new regression cases fail before the fix and pass after it.
- Full unit suite with Python 3.11 and the locked all-extras dependencies on Linux: **368 passed** (`pytest -m unit -n auto`, four workers).
- Ruff check, Ruff format check, pyrefly, the repository license check and Poetry build passed. Pyrefly reported zero errors with the existing suppressions and deprecation warnings.
- Filesystem suite: **19 passed**. Also checked regex semantics, glob/case options, result limits, no matches, a leading-hyphen relative root, cache/clone behavior and the standard-library fallback.


Tomas2D (approved):
Correct fix, and the test is well chosen — going through the public `GrepTool.run` API with real ripgrep, asserting the full match record rather than just a count, and marked `@pytest.mark.unit` so it actually runs under `pytest -m unit`. The `shutil.which("rg")` skip guard is the right call too.

Reproduced the behaviour you describe on ripgrep 15.2.0:

```
# pre-fix: rg --json --line-number --max-count=50 "--help" <root>
ripgrep 15.2.0
Andrew Gallant <jamslam@gmail.com>          <- help text, not a search

# post-fix: rg --json --line-number --max-count=50 -- "--help" <root>
"lines":{"text":"usage: app --help\n"},"line_number":1,...
  "submatches":[{"match":{"text":"--help"},"start":11,"end":17}]
```

**Worth knowing: this is doing more than fixing empty results.** `pattern` is model-controlled input being spliced into an `argv` ahead of any `--`, so it wasn't only `--help`/`--version` that got through — *any* ripgrep flag did. That includes `--pre=COMMAND`, which runs an arbitrary program over each searched file.

I couldn't turn that into execution, and I don't want to overstate it: because the tool passes exactly `[pattern, root]`, injecting a flag leaves a single positional, so ripgrep takes `root` as the *pattern* and reads from **stdin** rather than walking files — the preprocessor never gets a file. What it does reliably produce is a hang: `_run_ripgrep` calls `create_subprocess_exec` with `stdout`/`stderr` piped but **stdin left inherited**, so ripgrep blocks on the parent's stdin and `process.communicate()` never returns. My first attempt at this sat there until I killed it at 120s. In an agent run that's an unbounded stall from a crafted search pattern.

So I'd treat this as hardening as well as a correctness fix, and it makes the change more worth prioritising than the title suggests. The `--` also covers `input.root`, which is user-controlled and had the same exposure.

**One suggestion, separate from this PR:** consider passing `stdin=asyncio.subprocess.DEVNULL` in `_run_ripgrep`. That's defence in depth against exactly this class of stall, independent of argument parsing, and it's a one-line change. Happy for it to be a follow-up rather than scope creep here.

I've approved the workflow run so this gets CI coverage.

baba9811:
Thanks for the review and for reproducing the issue! I’ll look into isolating stdin with `DEVNULL` and a focused regression test as a separate follow-up.

PR #1661, opened 2026-09-17
fix(tools): isolate grep subprocess stdin

baba9811:

Follow-up to #1660, implementing the stdin isolation suggested in [this review](https://github.com/i-am-bee/beeai-framework/pull/1660#pullrequestreview-5235294247).

`GrepTool` searches files and exposes no stdin input, but its ripgrep subprocess inherits the host's stdin. The input schema accepts `root="-"`, which ripgrep interprets as stdin even after the option separator added in #1660. A public-API reproduction consumed the host's input; with that input stream held open, the tool waited for more data.

### Description

Pass `stdin=asyncio.subprocess.DEVNULL` when starting ripgrep so it cannot consume or wait on the host's inherited stdin. Argument construction, result parsing and the Python fallback are unchanged.

Add a unit regression that runs the public `GrepTool.run` API with real ripgrep in a child Python process. The test supplies known input to that process and verifies it remains available after the tool completes.

### Testing

- The regression fails before the change because ripgrep consumes the supplied input, and passes afterward.
- Full unit suite with Python 3.11 and the locked all-extras dependencies on Linux: **369 passed** (`pytest -m unit -n auto`, four workers).
- Filesystem suite: **20 passed**. Also verified completion while the host input pipe remains open, ordinary and `./-` directory searches, leading-hyphen patterns, case/glob options, result limits and the Python fallback.
- Ruff check, Ruff format check, pyrefly, the repository license check and Poetry build passed.


Tomas2D (approved):
This goes further than what I suggested, and the extra bit is the important part.

I framed the stall in #1660 as reachable through a crafted `pattern`, which the `--` separator then closed. **You found that it wasn't actually closed.** `root` is also caller-controlled, the schema accepts `root="-"`, and ripgrep treats a bare `-` as stdin *after* the separator as well. Confirmed against ripgrep 15.2.0:

```
$ echo "needle from stdin" | rg --json --line-number --max-count=50 -- "needle" -
"text":"needle from stdin\n"
```

So the option separator never addressed that path, and my review of #1660 was incomplete in implying it did. `stdin=DEVNULL` is the right fix precisely because it doesn't depend on argument parsing at all — it removes the capability rather than trying to block every route to it.

**The regression test is the part I'd single out.** Asserting "the tool didn't hang" is awkward and usually ends up as a timeout that passes vacuously. Running `GrepTool.run` inside a child Python process, feeding that process known stdin, and then asserting the input is *still readable afterwards* tests the actual property — that the subprocess didn't consume the parent's stream. It fails before the change and passes after, which is what makes it worth having. `check=True` plus `timeout=30` also means a regression shows up as a failure rather than a hung suite.

Everything else checks out: the `shutil.which("rg")` guard matches the surrounding tests, `@pytest.mark.unit` means it runs under `pytest -m unit`, and the modified call is ~115 characters against ruff's `line-length = 120`. Argument construction, result parsing and the stdlib fallback are untouched.

One note for the record, not a change request: with `DEVNULL` in place, `root="-"` now returns no matches instead of stalling, which is a sane failure mode. Explicitly rejecting `-` in the input schema would be clearer still, but it's cosmetic now that the capability is gone.

CI green. Merging.

PR #886, opened 2025-06-12
feat: add context dependent IO support

Tomas2D:
### Which issue(s) does this pull-request address?

Ref: #885

### Description

- add support for ACP / BeeAI Platform interception/async functionality (https://agentcommunicationprotocol.dev/how-to/await-external-response)
- wrap IO operations (currently just `input`) to be changed based on the execution environment (terminal = standard input, ACP = custom functionality, etc.)
- add `middlewares` argument for `AskPermissionRequirement`

#### General
- [ ] I have read the appropriate contributor guide: [Python](https://github.com/i-am-bee/beeai-framework/blob/main/CONTRIBUTING.md)
/ [TypeScript](https://github.com/i-am-bee/beeai-framework/blob/main/CONTRIBUTING.md)
- [ ] I have signed off on my commit: [Python instructions](https://github.com/i-am-bee/beeai-framework/blob/main/python/CONTRIBUTING.md#developer-certificate-of-origin-dco) / [TypeScript instructions](https://github.com/i-am-bee/beeai-framework/blob/main/typescript/CONTRIBUTING.md#developer-certificate-of-origin-dco)
- [ ] Commit messages and PR title follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary)
- [ ] Appropriate label(s) added to PR: `Python` for Python changes, `TypeScript` for TypeScript changes

#### Code quality checks
- [ ] Linting passes: Python `poe lint` or `poe lint --fix` / TypeScript `yarn lint` or `yarn lint:fix`
- [ ] Formatting is applied: Python `poe format` or `poe format --fix` / TypeScript: `yarn format` or `yarn format:fix`
- [ ] Static type checks pass: Python `poe type-check`

#### Testing
- [ ] Unit tests pass: Python `poe test --type unit` / TypeScript `yarn test:unit`
- [ ] E2E tests pass: Python `poe test --type e2e` / TypeScript: `yarn test:e2e`
- [ ] Integration tests pass: Python `poe test --type integration`
- [ ] Tests are included (for bug fixes or new features)

#### Documentation
- [ ] Documentation is updated
- [ ] Embedme embeds code examples in docs. To update after edits, run: Python `poe docs --type build`

PR #1662, opened 2026-09-18
fix(internals): restore default I/O in new contexts

baba9811:

The default file, shell and console backends are installed only in the context that imports their modules. Calling the public tools from a `ThreadPoolExecutor` worker therefore fails with `LookupError`. Importing them lazily inside an asyncio request also makes the first request succeed while a subsequent sibling request fails.

This prevents the documented local defaults from working without additional setup, even with valid file paths and commands.

### Description

Provide a fresh local fallback when reading each I/O ContextVar. Explicit backends and their setup/reset behavior remain unchanged, and no mutable object is shared as a constructor default, preserving the concern raised in [#886](https://github.com/i-am-bee/beeai-framework/pull/886#discussion_r2145264773).

Add ten regression cases covering file reads and both edit modes, subprocess output, console input and confirmation, context isolation, nested overrides and cleanup back to the local default. Update the ACP restoration test to verify public `io_read` behavior.

### Testing

- All ten new regression cases fail before the fix with the corresponding `LookupError` and pass after it. The focused filesystem, shell, console and ACP suites pass: **46 tests**.
- The original public-API reproductions pass for ordinary workers, sequential requests with lazy imports, the main context, `asyncio.to_thread`, explicit worker overrides, and console input in main/worker contexts.
- Full unit suite on Linux with Python 3.11 and the unchanged Poetry lock/all-extras dependencies: **380 passed** (`pytest -m unit -n auto`, four workers). The existing PyCharm diff-tools warning appears once per worker.
- Full Ruff check/format, pyrefly (**0 errors**), license check and Poetry sdist/wheel build pass. Pyrefly reports 307 suppressed diagnostics and existing deprecation warnings. The reviewed test-type correction was also checked with focused pyrefly and the console tests.
- Live editor and model-service E2E tests were not run.


Tomas2D (approved):
Reproduced the failure against current `main` before reviewing the fix:

```
ThreadPoolExecutor worker -> LookupError: <ContextVar name='io_storage'>
```

That confirms the diagnosis. `_storage.set(...)` at import time binds the value in whichever context happens to execute the import; a `ContextVar` set that way isn't visible from a different thread's context, nor from any context copied before the set. So the documented local defaults work in the importing context and nowhere else, and the lazy-import case you describe — first request fine, sibling request failing — falls straight out of the same mechanism.

Moving the default into `_storage.get(default)` is the right fix. It makes the fallback a property of every *read* rather than of one privileged context, so it can't be missed depending on who imported what and when. Constructing a fresh `IOHandlers` per call rather than a module-level instance also keeps the concern raised in [#886](https://github.com/i-am-bee/beeai-framework/pull/886#discussion_r2145264773) addressed — no mutable object shared as a default. The per-call allocation is trivial next to the awaited I/O.

Explicit backends and the setup/reset path are untouched, which is what I wanted to confirm: `get(default)` only supplies a value when nothing is set, so anyone who has called `set()` is unaffected.

The test coverage is appropriately broad for a context-propagation bug — worker threads, nested overrides, and cleanup back to the local default are exactly the cases where this class of bug hides, and testing through public `io_read` rather than the `ContextVar` keeps them meaningful if the storage mechanism changes. Updating the ACP test to assert public behaviour rather than internals is an improvement in its own right.

CI is green on your head commit.

One observation for a possible follow-up, not this PR: the two default handlers are now instantiated on every `io_read`/`io_confirm` call in the fallback path. That's fine, but if it ever matters, a module-level *immutable* sentinel would give the same isolation guarantee with one allocation. Not worth changing now — correctness first, and the current shape is the more obviously-correct one.

PR #1663, opened 2026-09-18
fix(backend): keep cached token counts when merging chunks

chrikrah:
`ChatModelOutput.merge` rebuilds `ChatModelUsage` from three of its five fields (`python/beeai_framework/backend/types.py:206-210`). It declares `cached_prompt_tokens` and `cached_creation_tokens` at `:61-66` and carries neither, so both reset to 0 on every merge.

`from_chunks` at `:181-186` is the only accumulator on the streaming path, reached from `chat.py:617`. A streamed response therefore always reports zero prompt-cache usage, while the same request with `stream=False` reports the real numbers.

```
usage after merge -> {'prompt_tokens': 120, 'completion_tokens': 30, 'total_tokens': 150,
                      'cached_prompt_tokens': 0, 'cached_creation_tokens': 0}
```

Both inputs carried `cached_prompt_tokens=96`.

## It is a regression

Before `ad3b1d6` this method did `merged_usage = self.usage.model_copy()`, which preserved the cached fields. `ad3b1d6`, "feat(agents): add cost and token usage info (#1334)", replaced that with a fresh three-field constructor, in the same change that introduced the two fields. They went from surviving to zeroed.

`5f13960` then applied a prompt-caching strategy to `RequirementAgent`. `ReActAgent` defaults to `stream: bool = True` (`react/agent.py:59`) and sums per-call totals through `usage.merge(it.raw.usage)` at `:217`, where `raw` comes from the streamed call. The framework has therefore been using caching it cannot measure.

## Why not reuse `ChatModelUsage.merge`

It adds; this path needs max. The two sit at different levels. `ChatModelUsage.merge` accumulates across separate model calls, at `requirement/_runner.py:120` and `react/agent.py:217`. `ChatModelOutput.merge` accumulates chunks of one response, each carrying a snapshot of the same totals, so summing double-counts:

```
SUM semantics (ChatModelUsage.merge): prompt_tokens=240 total_tokens=270 cached_prompt_tokens=192
MAX semantics (this change)         : prompt_tokens=120 total_tokens=150 cached_prompt_tokens=96
```

The TypeScript twin agrees: `typescript/src/backend/chat.ts:585-604` merges every field with `takeBigger`, including `cachedPromptTokens` at `:598`.

The change is two `max(...)` lines beside the three already there.

## Tests

No test anywhere covered this accumulator. `from_chunks` and `merge` had zero test references before this. The new file carries two cases. The first merges two outputs whose cached values are both non-zero and different, which pins per-field max rather than "take the other side". The second goes through `from_chunks`, mirroring the real streaming path.

```
$ pytest tests/backend/ -m unit -n auto
86 passed

$ pytest -m unit -n auto
372 passed
```

Revert the production change, keep the test, and both cases fail with `assert 0 == 96`.

`ruff check` passes, `ruff format --check` reports 613 files already formatted, `pyrefly check` reports 0 errors, and `cz check --rev-range origin/main..HEAD` validates.

## Not checked

`addlicense -check` did not run, because the binary is mise-managed and mise is absent here. A textual comparison against `test_utils.py` covered the license header instead. End-to-end tests did not run either, since they need a local Ollama with `llama3.1:8b` and `granite4:latest`.

#909, "Usage when streaming is always empty", is a closed precedent rather than a duplicate. Its cause was that nothing requested usage from the provider, which `dec59f5` fixed in the adapter with `include_usage`, still live at `litellm/chat.py:150`. That lands five months before `ad3b1d6` created this one. Same symptom family, different cause, non-overlapping fixes.

No open issue covers this, so the template's `Closes:` line has nothing to point at. None of the nine open pull requests touches `backend/types.py`.


Tomas2D (approved):
Checked the reasoning rather than just the diff, and it holds up. This is a well-argued PR.

**The defect.** `ChatModelUsage` declares five fields; `ChatModelOutput.merge` rebuilds it from three, so `cached_prompt_tokens` and `cached_creation_tokens` reset to `0` on every merge. Since `from_chunks` is the only accumulator on the streaming path, a streamed response reports zero prompt-cache usage while the identical non-streamed request reports the real numbers.

**Max, not sum, is right here** — and this is the part I wanted to confirm independently, because the two `merge` methods are easy to conflate. They operate at different levels: `ChatModelUsage.merge` accumulates across separate model calls and correctly uses `+=`, while `ChatModelOutput.merge` folds chunks of a *single* response where each chunk carries a snapshot of the same running totals. Summing snapshots double-counts. The three existing fields already use `max`, as does the `cost` block immediately above, so the two new lines are consistent with their surroundings rather than introducing a new convention.

**The cross-language check confirms it.** `typescript/src/backend/chat.ts:585-604` merges every usage field with `takeBigger`, `cachedPromptTokens` included:

```ts
this.usage = customMerge([this.usage, other.usage], {
  totalTokens: takeBigger,
  promptTokens: takeBigger,
  completionTokens: takeBigger,
  cachedPromptTokens: takeBigger,
  reasoningTokens: takeBigger,
});
```

So Python was the outlier, not TypeScript.

**On the test.** Using two different non-zero cached values on both sides is the right call — it pins per-field `max` rather than "whichever side is populated", which a `0` on one side would let pass either way. Good that you noted reverting the production change makes both cases fail with `assert 0 == 96`; that's the check I'd have asked for otherwise.

Two small notes, neither blocking:

- Incidental to this PR, but the two languages have drifted in their *field sets*: TypeScript tracks `reasoningTokens` which Python's `ChatModelUsage` lacks, and Python tracks `cached_creation_tokens` which TypeScript lacks. Worth a separate issue if anyone cares about parity.
- On `addlicense` not having run: I checked the new file's header against the repo's and it matches what the license check expects, so that's covered.

Workflow run approved. Once it's green this is good to go.

Issue #1233, opened 2025-10-26
Pydantic validation error in ChatToolFunctionDefinition when parsing tool calls from watsonx Orchestrate


jaimcuev:
**Describe the bug**
When consuming BeeAI agents through the `WatsonxOrchestrateServer` and calling them from IBM watsonx Orchestrate, the server fails to parse tool calls because the `arguments` field comes as a JSON string instead of a dict object. The Pydantic model `ChatToolFunctionDefinition` has a field_validator with incorrect decorator ordering, causing the validation to fail.

**To Reproduce**
1. Create and start a WatsonxOrchestrateServer with an agent that uses tools:
```python
from dotenv import load_dotenv

from beeai_framework.adapters.watsonx_orchestrate import WatsonxOrchestrateServer, WatsonxOrchestrateServerConfig
from beeai_framework.agents.requirement import RequirementAgent
from beeai_framework.backend import ChatModel
from beeai_framework.serve.utils import LRUMemoryManager

load_dotenv()

def main() -> None:
    llm = ChatModel.from_name("watsonx:meta-llama/llama-4-maverick-17b-128e-instruct-fp8")
    agent = RequirementAgent(
        llm=llm,
        role="friendly AI assistant",
        instructions="Be helpful and conversational in all your interactions."
    )

    config = WatsonxOrchestrateServerConfig(port=8080, host="0.0.0.0", api_key=None)
    server = WatsonxOrchestrateServer(config=config, memory_manager=LRUMemoryManager(maxsize=100))
    server.register(agent)

    server.serve()
    
if __name__ == "__main__":
    main()
```
2. Register the agent in IBM watsonx Orchestrate UI or CLI
3. Send a message from watsonx Orchestrate
4. See Pydantic validation error when the server tries to parse the `arguments` field

**Expected behavior**
The server should correctly parse tool calls where the `arguments` field is a JSON string and convert it to a dict object without raising validation errors.

Example request received from watsonx Orchestrate:
```json
{
    "stream": true,
    "messages": [
        {
            "content": "Today's date is Sun Oct 26, 2025.\n\nYou are news_agent, a helpful, ethical, and cautious AI assistant. You carefully follow instructions. You are helpful and harmless and you follow ethical guidelines and promote positive behavior. \nTHIS IS VERY IMPORTANT - you will follow these instructions very carefully: \nYou should try to answer the user's questions to the best of your ability..\n\n\n",
            "role": "system"
        },
        {
[...]

PR #1664, opened 2026-09-18
fix(internals): register the two field validators the decorator order disabled

chrikrah:
Two classes put `@classmethod` above `@field_validator`. Pydantic collects a validator by the marker `field_validator` leaves on the class attribute, and `classmethod` wraps that marker out of reach, so neither validator was ever registered. On pydantic 2.12.5 both classes report an empty validator set:

```
openmeteo validators: []
chat validators: []
```

## `OpenMeteoToolInput._to_lower`

`python/beeai_framework/tools/weather/openmeteo.py:36-38`. It exists to accept a `temperature_unit` in any case, and `1ddf343`, the commit that added it, is titled *"feat(tools): ignore case sensitivity for OpenMeteoTool's temperature_unit parameter"*. The field is a `Literal["celsius", "fahrenheit"]`, so with the validator dead:

```
OpenMeteoToolInput(location_name="Berlin", temperature_unit="CELSIUS")
ValidationError: Input should be 'celsius' or 'fahrenheit'
  [type=literal_error, input_value='CELSIUS', input_type=str]
```

A model is as likely to emit `"Celsius"` as `"celsius"`, which is the case the commit set out to handle.

## `ChatToolFunctionDefinition.parse_arguments`

`python/beeai_framework/adapters/openai/serve/chat_completion/_types.py:16-18`. This is the worse of the two. It exists to decode tool-call `arguments` that arrive as a JSON string, which is how the OpenAI API sends them, and it calls `parse_broken_json`, so it was written for malformed model output as well. Unregistered:

```
ChatToolFunctionDefinition(name="f", arguments='{"a": 1}')
ValidationError: Input should be a valid dictionary
  [type=dict_type, input_value='{"a": 1}', input_type=str]
```

That is the OpenAI-compatible serve path rejecting the standard wire format.

## The fix

Swapping the two decorators registers both:

```
openmeteo validators: ['_to_lower']     CELSIUS -> celsius
chat validators: ['parse_arguments']    '{"a": 1}' -> {'a': 1}
```

These are the same mistake in two places, so they are one change.

## This defect class is already accepted here

#1233, "Pydantic validation error in ChatToolFunctionDefinition when parsing tool calls from watsonx Orchestrate", reported exactly this, and #1234 fixed it on 2025-10-27. That pull request touched one file, `adapters/watsonx_orchestrate/_api.py`, and its own description says "The `@field_validator` decorator must come before".

The two sites here are different files and were not part of it. `ChatToolFunctionDefinition` exists in both the watsonx Orchestrate adapter and the OpenAI serve adapter, so the name collides across a search.

After this change no site remains. `git grep -B1 '@field_validator'` and the same for `@model_validator` over `python/` return no other `@classmethod` on the line above.

## Tests

Four unit tests. Two pin registration and fail against the unmodified files. The other two guard against over-correction: a dict passes through unchanged, and unparseable arguments still raise.

```
$ pytest -m unit -n auto
374 passed in 36.39s
```

Revert both production files, keep the tests, and the two registration tests fail while the two guards still pass.

`ruff check` and `ruff format --check` are clean. `pyrefly check` reports 0 errors with 304 suppressed.

## Not checked

No live Open-Meteo or OpenAI call. Both defects are in pydantic model construction, which the reproductions drive directly.

`pyrefly` initially reported 619 errors in the worktree, which is the missing virtualenv rather than the change: run in a tree with dependencies installed it is 0.

No open issue covers this, so the template's `Closes:` line has nothing to point at.


Tomas2D (approved):
Verified both, independently of the writeup. Confirmed on pydantic 2.12.5 against current `main`:

```
OpenMeteoToolInput.__pydantic_decorators__.field_validators -> []
OpenMeteoToolInput(location_name='Berlin', temperature_unit='CELSIUS') -> ValidationError

ChatToolFunctionDefinition.__pydantic_decorators__.field_validators -> []
ChatToolFunctionDefinition(name='f', arguments='{"a": 1}')            -> ValidationError
```

Both validator sets are genuinely empty, so neither function has ever run. Swapping the decorators so `@field_validator` is outermost is the correct fix — pydantic reads the marker off the class attribute, and `classmethod` wraps it out of reach.

**I'd rank the second one as the more serious of the two.** `1ddf343` made `temperature_unit` case-insensitive and that has silently not worked since it landed, which is annoying but degrades to a validation error on a single optional field. `ChatToolFunctionDefinition.parse_arguments`, though, exists because the OpenAI API sends tool-call arguments as a JSON *string* — so on the `OpenAIServer` chat-completion path, any genuine tool call from an OpenAI-compatible client is rejected outright. That's a functional break in the serve adapter, not a papercut, and it's worth calling out in the changelog as such.

**I also checked whether this pattern appears anywhere else.** It doesn't — there are five validator decorators in `beeai_framework/`, and the only two with `@classmethod` on the outside are the two you've fixed. So this PR is complete rather than a sample.

Tests look right: they assert the behaviour the validators exist to provide (a JSON string decoding, a dict passing through unchanged, invalid JSON raising) rather than poking at `__pydantic_decorators__`, which means they'd survive an internals change in pydantic. All marked `@pytest.mark.unit`, so they run under `pytest -m unit`.

One thought, not a change request: a validator that silently does nothing is invisible precisely because the class still constructs fine. If you wanted belt-and-braces against a recurrence, a small test asserting each model's `field_validators` is non-empty would catch a future re-ordering directly — but the behavioural tests you've written already fail if it regresses, so this is optional.

Workflow run approved.

Tomas2D:
CI caught something the local checks wouldn't have: `pyrefly` rejects the test itself.

```
ERROR Argument `Literal['{"a": 1}']` is not assignable to parameter `arguments`
      with type `dict[Any, Any]` [bad-argument-type]
```

Which is a fair complaint, and a slightly awkward consequence of the thing you're fixing: `arguments` is annotated `dict[str, Any]` because that's the *output* type, while `mode="before"` is what makes the string acceptable at runtime. A static checker only sees the annotation, so the constructor call looks wrong even though it works.

I've pushed `1b797b16` to your branch switching the three cases to `model_validate`:

```py
ChatToolFunctionDefinition.model_validate({"name": "f", "arguments": '{"a": 1}'})
```

`model_validate` takes `Any`, so the checker is satisfied without a `# pyrefly: ignore` suppression — and I'd argue it's the better test regardless, because it's how these objects are actually built in the serve path: from a decoded request body, not from keyword arguments. So the test now exercises the route the validator was written for.

Verified locally: 3 passed, `ruff check` clean, `ruff format --check` clean across 613 files, and `pyrefly` reports nothing for either `chat_completion` or `openmeteo`.

My approval stands — the production change was always correct, this was only the test's type surface. The workflow should re-run on the new head.

PR #1666, opened 2026-09-18
fix(tools): parse a YAML OpenAPI schema with yaml.safe_load, not yaml.parse

chrikrah:

There is no issue for this. The template offers the alternative of describing the problem here, so the description below does that.

### Description

`OpenAPITool.from_url` calls `yaml.parse(response.text)` at `python/beeai_framework/tools/openapi.py:246` when the server's `Content-Type` names YAML.

`yaml.parse` is the low-level event API. It yields `Event` objects and never returns the document. Line 255 then calls `.get` on that generator:

```
AttributeError: 'generator' object has no attribute 'get'
  beeai_framework/tools/openapi.py:255
```

So every YAML-served OpenAPI schema fails, and `from_url` handles JSON only. `yaml.safe_load` returns the parsed mapping.

One line. The change leaves the JSON branch alone.

### Tests

`python/tests/tools/test_openapi.py` is new, since the module had no tests.

It serves a minimal YAML schema through a patched `httpx.AsyncClient.get`, then asserts that `from_url` builds one tool from the schema's single operation.

```
$ .venv/bin/python -m pytest tests/tools -m unit -q
50 passed, 13 deselected

$ .venv/bin/python -m ruff check --quiet beeai_framework/tools/openapi.py tests/tools/test_openapi.py
$ .venv/bin/python -m ruff format --quiet --check beeai_framework/tools/openapi.py tests/tools/test_openapi.py
```

Revert `openapi.py` alone, keep the test, and the test fails with that same `AttributeError`.

#### General

#### Code quality checks

#### Testing
  `python:test:e2e` needs an Ollama instance serving `llama3.1:8b` and `granite4:latest`, per `python/CONTRIBUTING.md`. I did not run it. The change is one function call in a YAML branch, and the new unit test covers that call.

#### Documentation
  No document names the PyYAML function that parses the schema.


Tomas2D (approved):
Confirmed against current `main`:

```
yaml.parse(...)  -> generator
  .get('paths')  -> AttributeError: 'generator' object has no attribute 'get'
yaml.safe_load(...) -> {'openapi': '3.0.0', 'info': {'title': 't'}, 'paths': {}}
```

`yaml.parse` is the low-level event API, so the YAML branch of `from_url` has never worked — `from_schema` receives a generator and dies on the first `.get`. One-line fix, correct function.

`safe_load` rather than `load` is also the right choice on its own merits: the schema comes from a remote URL, and `yaml.load` without a safe loader can instantiate arbitrary Python objects from a crafted document. Worth noting that this path is fetching untrusted-ish content, so `safe_load` isn't just convention here.

Good that the test is the first coverage this module has had, and that it goes through `from_url` with a patched transport rather than calling `from_schema` directly — that's what makes it exercise the broken branch. Reverting `openapi.py` and keeping the test reproduces the `AttributeError`, as you say.

On the checklist items you couldn't complete: the label needs write access so that's expected, and running `ruff check` / `ruff format --check` directly is exactly what `python:check` does, so nothing was skipped in substance. CI is green.

PR #1667, opened 2026-09-18
fix(internals): apply the default retry backoff factor when the config leaves it unset

chrikrah:

No issue was raised. The template offers the alternative of describing the problem here, so the description below does that.

### Description

`do_retry` computes the backoff at `python/beeai_framework/retryable.py:56`:

```python
factor = options.get("factor", 2) or 0
```

Its only caller, `Retryable.get`, always sends the key, at `:144`:

```python
options = {"factor": self._config.factor, ...}
```

The key is therefore never absent, so the `.get` default of `2` can never fire. `RetryableConfig.factor` defaults to `None`, and `None or 0` is `0`. Any caller that does not set `factor` retries with no backoff at all: `asyncio.sleep(0 ** (attempt - 1))`.

Three call sites leave it unset, so all three retry back to back:

- `backend/embedding.py:197`
- `agents/react/runners/default/runner.py:227`
- `agents/react/runners/default/runner.py:301`

Two set it deliberately and are unaffected: `tools/tool.py:177` passes `or 1`, and `backend/chat.py:544` passes `factor=0` for its invalid-tool-call loop.

The TypeScript sibling has the intended default, at `typescript/src/internals/helpers/retry.ts:29`:

```ts
const factor = options?.factor ?? 2;
```

The change replaces only `None`, so an explicit `0` still means no backoff and `chat.py`'s choice survives.

### Tests

```
$ .venv/bin/python -m pytest tests/test_retryable.py tests/backend/ tests/agents/ -m unit -q
101 passed, 3 deselected

$ .venv/bin/python -m ruff check --quiet beeai_framework/retryable.py tests/test_retryable.py
$ .venv/bin/python -m ruff format --quiet --check beeai_framework/retryable.py tests/test_retryable.py
```

Two new cases: one asserting the unset default now backs off `[2, 4]`, one pinning that `factor=0` still yields `[0, 0]`. Both mock `asyncio.sleep`, so neither waits.

Revert `retryable.py` alone, keep them, and the first fails with `assert [0, 0] == [2, 4]`.

One note on the existing suite: `test_retryable_retries` does not mock `asyncio.sleep` and does not pass `factor`, so it now waits out a real two-second backoff on its second attempt. It still passes. I left it alone rather than fold an unrelated edit into this change.

#### General

#### Code quality checks

#### Testing
  `python:test:e2e` needs an Ollama instance serving `llama3.1:8b` and `granite4:latest`, per `python/CONTRIBUTING.md`. I did not run it. The change is a two-line default in a pure function, and the two new unit tests cover both branches of it.

#### Documentation
  No document states a default backoff factor for `RetryableConfig`.


Tomas2D (approved):
Verified the collapse against current `main`:

```
RetryableConfig(max_retries=3).factor = None
computed factor                       = 0
sleep for attempts 2,3,4              -> [0, 0, 0]
if the intended default had applied   -> [2, 4, 8]
```

So `options.get("factor", 2)` really is unreachable as a default — `Retryable.get()` always sends the key, and `None or 0` collapses to `0`. Every caller that leaves `factor` unset retries with no delay whatsoever.

That's worse than a cosmetic default being wrong, because of *which* callers are affected. `backend/embedding.py` and both ReAct runner sites are precisely the paths that retry against a remote model provider — the place exponential backoff exists to protect. Retrying a rate-limited or overloaded endpoint three times back-to-back is close to the opposite of the intended behaviour, and it would present as "the provider keeps 429-ing us" rather than as a framework bug, which is probably why it survived this long.

**The fix handles the distinction that matters.** Replacing only `None` and leaving an explicit `0` intact preserves `backend/chat.py:547`, which passes `factor=0` deliberately for the invalid-tool-call loop — that one *wants* no delay. A naive `or 2` would have silently introduced backoff there. The comment explaining why `.get("factor", 2)` can't express this is worth having; it's the kind of thing that gets "simplified" back into a bug later.

Mirroring the TypeScript `?? 2` also resolves a real cross-language divergence rather than inventing a number.

Test looks right: monkeypatching `asyncio.sleep` to record delays asserts the actual observable behaviour rather than reaching into the options dict, and it would fail if someone reverted to the `or 0` form.

CI green. Merging.

PR #1668, opened 2026-09-18
fix(tools): stop reporting a full result set as truncated in GrepTool

chrikrah:

There is no issue for this. The template offers the alternative of describing the problem here, and the description below does that.

### Description

Both search paths in `GrepTool` derive `truncated` from a comparison that holds when the match count merely reaches the cap, at `grep_tool.py:98` and `:129`:

```python
truncated = len(matches) >= input.max_results
```

A search that found every match there is, and happened to find exactly `max_results` of them, then tells the caller more remains. For an agent deciding whether to narrow its pattern or page again, that decides between stopping and looping.

Measured on both paths, three files each holding one `hello`, with `max_results=3`:

```
exact fit      files=3 cap=3 -> returned=3 truncated=True    before
exact fit      files=3 cap=3 -> returned=3 truncated=False   after
really more    files=5 cap=3 -> returned=3 truncated=True    unchanged
under the cap  files=2 cap=3 -> returned=2 truncated=False   unchanged
```

The ripgrep path now asks `rg` for one more than the cap. Both paths break at more-than rather than at-least, then slice back to `max_results`. A result set genuinely larger than the cap still reports `truncated`.

### Tests

```
$ .venv/bin/python -m pytest tests/tools -m unit -q
59 passed, 13 deselected

$ .venv/bin/python -m ruff check --quiet beeai_framework/tools/filesystem/grep_tool.py tests/tools/filesystem/test_grep.py
$ .venv/bin/python -m ruff format --quiet --check beeai_framework/tools/filesystem/grep_tool.py tests/tools/filesystem/test_grep.py
```

Three new cases: the exact-fit boundary on each path, and one pinning that exceeding the cap still reports truncated.

The existing `fs_tree` fixture holds exactly two case-sensitive `hello` matches, so a cap of two hits the boundary without a new fixture.

Revert `grep_tool.py` alone, keep them, and `tests/tools/filesystem` gives `2 failed, 25 passed`.

#### General

#### Code quality checks

#### Testing
  `python:test:e2e` needs an Ollama instance serving `llama3.1:8b` and `granite4:latest`, per `python/CONTRIBUTING.md`. I did not run it. This tool makes no model call, and the three new unit tests cover both of its paths.

#### Documentation
  No prose describes the `truncated` field beyond the tool's own schema, which is unchanged.


Tomas2D (approved):
Correct fix, and the measurement table in the description is exactly the right way to present it — the three cases (exact fit, genuinely more, under the cap) are the full truth table for this predicate, and only the first one changes.

`len(matches) >= input.max_results` can't distinguish "found everything, and it happened to be exactly the cap" from "hit the cap and stopped", because at that point both look identical. Asking `rg` for `max_results + 1` and then slicing back is the standard way out: it buys exactly the one extra data point needed to tell those apart, at the cost of one additional match being parsed. Breaking on more-than rather than at-least in the stdlib path does the same thing without needing the extra fetch, since that loop controls its own iteration.

Worth stating why this matters beyond tidiness, since a stray boolean sounds cosmetic: `truncated` is what an agent reads to decide whether to narrow its pattern or page again. A false `True` on a complete result set sends it round a loop that can only ever return the same matches — so the cost is a wasted tool call and possibly a needlessly narrowed search, not just a wrong field.

Both paths fixed consistently, which is the part I'd have checked first — it would be easy to fix the ripgrep path and leave the fallback reporting the old answer, and then the behaviour would silently depend on whether `rg` happens to be installed.

CI green. Merging.

PR #1670, opened 2026-09-20
fix(emitter): clean up child pipe listeners on destroy to prevent mem…

saitejagirada:
## Problem

`Emitter.child()` pipes the new child emitter into its parent and appends the
pipe-cleanup closure to the **parent's** `_cleanups` list:

```python
cleanup = child_emitter.pipe(self)
self._cleanups.append(cleanup)
```

When the child is later destroyed (e.g. by `RunContext.enter()`'s `finally`
block at `context.py:272`), `child.destroy()` clears the **child's** listeners
and cleanups — but the stale closure remains in the **parent's** `_cleanups`.

Every `agent.run()` call creates a new `RunContext` → new child emitter → one
more dead closure accumulating on the long-lived agent emitter. After *N* runs,
`agent.emitter._cleanups` holds *N* orphaned entries that also prevent the
child `Emitter` objects from being garbage-collected.

**Measured impact (10 sequential `agent.run()` calls on one agent instance):**

| Metric | Before fix | After fix |
|---|---|---|
| `agent.emitter._cleanups` length | grows 1→10 | stays at 1 |
| Retained `Emitter` objects after `gc.collect()` | 12 | 2 |

## Fix

One-line addition in `Emitter.child()`: register a reverse-cleanup on the
**child** so that when `child.destroy()` runs, it removes the now-dead pipe
closure from the parent's `_cleanups`:

```python
child_emitter._cleanups.append(
    lambda: self._cleanups.remove(cleanup) if cleanup in self._cleanups else None
)
```

The `if cleanup in self._cleanups` guard handles the case where the parent is
destroyed first (its `_cleanups.clear()` already removed the entry).

No changes to `destroy()` or any other method. No changes outside `emitter.py`.

## Tests

Three new unit tests in `test_emitter.py`:

- `test_child_emitter_destroy_cleans_up_parent_reference` — single child destroy removes parent cleanup
- `test_multiple_child_emitters_do_not_leak_cleanups` — 10 create/destroy cycles, parent stays clean
- `test_parent_destroy_cleans_up_all_children` — parent destroy clears everything, children can still safely destroy after


Tomas2D (approved):
Reproduced the leak on current `main` and verified the fix eliminates it:

```
UNFIXED  parent cleanups: base=0  after 10 create/destroy cycles=10  leaked=10
FIXED    parent cleanups: base=1  after 10 create/destroy cycles=1   leaked=0
```

The asymmetry is real: `child()` appends the pipe-cleanup to the *parent's* `_cleanups`, but `destroy()` only ever clears the list belonging to the emitter being destroyed. So the closure outlives the child that justified it. Since `RunContext.enter()` creates a child emitter per run and destroys it in its `finally`, a long-lived agent accumulates exactly one dead closure per `agent.run()` — which matches your measurement.

Registering the reverse cleanup on the child is the right shape: the entry is removed by whichever side is torn down first, and the `if cleanup in self._cleanups` guard covers parent-destroyed-first, where `_cleanups.clear()` has already dropped it. I checked that ordering explicitly and it's a clean no-op.

I also want to flag that I initially misread `test_parent_destroy_cleans_up_all_children`. `assert len(parent._cleanups) == 3` looked wrong to me, since an unfixed parent holds 2 after two children. It's correct: with the fix, `Emitter.root().child(...)` registers the reverse cleanup *on the returned emitter*, so `parent` starts at 1 rather than 0, and your `# root's child + 2 children` comment says exactly that. Worth noting only because the assertion encodes a subtlety that a future reader may well trip over the same way — it's the one place the test depends on the fix's own side effect.

No mutation-during-iteration hazard either, which was my other worry: `destroy()` iterates the *child's* list while the new lambda mutates the *parent's*, so the two never alias.

Two observations, neither blocking:

- `list.remove()` is O(n) in the parent's cleanup count, so tearing down N simultaneously-live children is O(N²). Irrelevant for the agent pattern, where each child is destroyed promptly and N stays ~1, but worth knowing if anything ever fans out a large number of concurrent children from one parent.
- The closure now holds a strong reference from child to parent, where previously the reference ran only parent → child. Python's cycle collector handles that, and your 12 → 2 retained-object measurement shows it comes out ahead in practice — just noting the direction changed.

CI green. Merging.

PR #1671, opened 2026-09-20
fix(tools): keep extend metadata overrides independent

mohammedmessaoudene-cmd:
### Which issue(s) does this pull request address?

No prior issue. This PR fixes a self-contained metadata mapping bug in `Tool.extend`: the extended description was read from `overrides.name` instead of `overrides.description`.

### Description

- use `overrides.description` when constructing an extended tool
- add table-driven coverage for absent, empty, name-only, description-only, and combined overrides
- verify the input schema, mapped execution, and source tool remain unchanged

### Validation

- focused test: 30/30 passed
- TypeScript unit suite under `src`: 45 files, 337/337 passed
- TypeScript typecheck: passed
- ESLint and Prettier on the affected files: passed
- `git diff --check`: passed

The examples/E2E path was not used as a release gate because it invokes external local services such as Ollama; those services were not configured for this focused fix.

### Checklist

- [x] I have read the TypeScript contributor guide.
- [x] I have signed off on my commit.
- [x] The commit message and PR title follow Conventional Commits.
- [x] Code quality checks pass for the affected code.
- [x] Unit tests pass and regression coverage is included.
- [ ] E2E tests were not run; this metadata-only fix is covered by the unit suite and E2E requires external services.
- [x] No documentation change is needed for this internal metadata correction.

AI assistance was used to identify, test, and review this change. I reviewed the diff and validation results before submission.


Tomas2D (approved):
Confirmed against current `main`. Built a tool and called `extend()` both ways:

```
override name only  -> name: new-name      | description: new-name
override desc only  -> name: original-name | description: original description
```

So it's two symptoms from the one typo, and the second is the one I'd lead with:

1. `extend({ name })` silently **overwrites the description with the name**.
2. `extend({ description })` is **ignored entirely** — the description override has never worked at all.

That's worth calling out as more than a metadata nit. A tool's `description` is the text the model reads to decide whether to call it. Renaming a tool through `extend()` therefore replaces a written explanation of what the tool does with a bare identifier, which degrades tool selection at runtime rather than just looking wrong in a dump. And anyone who *tried* to fix that by passing `description` found it had no effect, with no error to explain why.

The fix is the obvious one-character-class change, and `overrides?.description || this.description` keeps the existing fallback semantics for absent/empty values, matching the `name` line directly above it.

Test coverage is well chosen: the table-driven cases (absent, empty, name-only, description-only, combined) pin exactly the matrix where this went wrong, and asserting that the source tool is unchanged guards against `extend()` mutating its parent. Empty-string is the case I'd have specifically asked for, since `||` treats it as absent — good that it's pinned rather than left to inference.

I've approved the workflow run so this gets CI coverage. Thanks also for stating the AI assistance and what you verified yourself — that's the right way to disclose it.

PR #1674, opened 2026-09-20
fix(python): forward-compat fixes for Python 3.14

### LINKED ISSUE #1588: Python 3.14 support

naikvaib:
Relates to #1588 (Python 3.14 support) — this is a partial slice, not a full close.

The maintainer's own analysis in the issue thread concluded that relaxing `requires-python`/the poetry `litellm` bound is not actionable right now — 3.14 support is blocked upstream on litellm — and recommended tracking upstream rather than editing the constraint locally. I left `python/pyproject.toml`'s `requires-python = ">=3.11,<3.14"` and `litellm = "^1.84.0"` untouched, matching that decision.

That same thread flagged two concrete, narrow forward-compat bugs unrelated to the litellm blocker, which this PR fixes:

1. `utils/asynchronous.py`: `asyncio.iscoroutinefunction` is deprecated (removal in 3.16) — switched to the already-imported `inspect.iscoroutinefunction`, its documented replacement.
2. `tests/utils/test_json_schema.py::test_json_schema_model`: three assertions compared `str(annotation) == "typing.Optional[...]"`; Python 3.14 renders `Optional[X]` as PEP 604 `X | None`, breaking the string comparison. Changed to compare the type object directly (`== X | NoneType`), matching the pattern already used elsewhere in the same file.

Verified against a real CPython 3.14.6 install (built a minimal venv with only the deps actually needed, since full `poetry install` is blocked by the same `<3.14` bound) — reproduced both original failures, confirmed the fix resolves them. `tests/utils/` (57 tests) and `tests/test_emitter.py` (24 tests) pass on 3.14. `ruff check`/`ruff format --check` clean.

This does not fully close #1588 — the `requires-python`/litellm bound is still upstream-gated.


Tomas2D (commented):
This is the right way to slice #1588 — thanks for reading the thread and *not* touching the constraint. The `requires-python` / litellm bound genuinely is upstream-gated, and #1599 is currently stuck on exactly that, so carving out the two independent forward-compat bugs is the part that can actually land.

I verified both.

**1. `asyncio.iscoroutinefunction` → `inspect.iscoroutinefunction`.** Correct substitution, and `inspect` is already imported in that module so the diff is genuinely one line. The thing I wanted to check before endorsing it is that the two don't disagree in practice, since they historically diverged. On 3.11 they agree across every case that matters here:

```
coroutine fn      asyncio=True   inspect=True
plain fn          asyncio=False  inspect=False
functools.partial asyncio=True   inspect=True
AsyncMock         asyncio=True   inspect=True
MagicMock         asyncio=False  inspect=False
object w/ async __call__  both False
```

`AsyncMock` was the one I expected to bite — it's the classic divergence — and it doesn't. Worth noting the failure mode would have been mild anyway: `ensure_async` only uses the result to decide whether to wrap in `asyncio.to_thread`, and the wrapper re-checks with `inspect.isawaitable`, so a false negative would be wasteful rather than wrong.

**2. The `typing.Optional[...]` string assertions.** Real 3.14 breakage — PEP 604 rendering changes `Optional[X]` to `X | None`, so a string comparison can't survive it. Comparing the type object is both version-independent and consistent with the `X | NoneType` pattern already used elsewhere in that same file, which is the detail that makes it a cleanup rather than a workaround.

**One blocker: DCO is red.** Commit `7b76165` has no `Signed-off-by` trailer. `git commit --amend --signoff && git push --force-with-lease` clears it.

I've approved the workflow run so you get CI signal on the current head. Once DCO is green and you flip this out of draft, I've no further requests — and it's worth landing independently of #1588 rather than waiting for the upstream unblock, since both fixes are correct on 3.11 today.

Building a minimal 3.14 venv to reproduce the failures first, rather than inferring them, is the right approach given `poetry install` is blocked by the very bound under discussion.

